PR #1, opened 2016-06-02
Creating a catalog is not supported. Clarify HTTP response

sheepkiller:
Hi,

I spent some time to  create a catalog, until I noticed it wasn't supported. I think the response should be more explicit than "null".
This one line patch clarify the status of the API, but I'm wondering if a 501 status code would be better than a 415.

Thanks.


ajoymajumdar:
Thanks. We have combined both unsupported media type and operation not supported under 415. Probably because 501 was not one of the available Status enumerations in jsr311. We should separate out the status response.

PR #17, opened 2016-12-02
Adding a notification service with a SNS implementation

tgianos:
Implementations receive internal events and can publish external notifications based on implementations of the interfaces.

### DIFF
--- a/metacat-common/src/main/java/com/netflix/metacat/common/MetacatRequestContext.java
+++ b/metacat-common/src/main/java/com/netflix/metacat/common/MetacatRequestContext.java
@@ -19,6 +19,9 @@ package com.netflix.metacat.common;
 
 import lombok.Data;
 
+import java.util.Date;
+import java.util.UUID;
+
 /**
  * The context of the request to Metacat.
  *
@@ -44,6 +47,9 @@ public class MetacatRequestContext {
      */
     public static final String HEADER_KEY_DATA_TYPE_CONTEXT = "X-Netflix.data.type.context";
 
+    private final String id = UUID.randomUUID().toString();
+    // TODO: Move to Java 8 and use java.time.Instant
+    private final long timestamp = new Date().getTime();
     private final String userName;
     private final String clientAppName;
     private final String clientId;
--- a/metacat-common/src/main/java/com/netflix/metacat/common/dto/notifications/sns/SNSMessageFactory.java
+++ b/metacat-common/src/main/java/com/netflix/metacat/common/dto/notifications/sns/SNSMessageFactory.java
@@ -1,0 +1,83 @@
+/*
+ *
+ *  Copyright 2016 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.metacat.common.dto.notifications.sns;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.metacat.common.dto.notifications.sns.messages.AddPartitionMessage;
+import com.netflix.metacat.common.dto.notifications.sns.messages.CreateTableMessage;
+import com.netflix.metacat.common.dto.notifications.sns.messages.DeletePartitionMessage;
+import com.netflix.metacat.common.dto.notifications.sns.messages.DeleteTableMessage;
+import com.netflix.metacat.common.dto.notifications.sns.messages.UpdateTableMessage;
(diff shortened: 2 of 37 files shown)


metacat-common/src/main/java/com/netflix/metacat/common/MetacatRequestContext.java
```diff
@@ -32,6 +35,9 @@
     public static final String HEADER_KEY_JOB_ID = "X-Netflix.job.id";
     public static final String HEADER_KEY_DATA_TYPE_CONTEXT = "X-Netflix.data.type.context";
 
+    private final String id = UUID.randomUUID().toString();
+    // TODO: Move to Java 8 and use java.time.Instant
+    private final long timestamp = new Date().getTime();
```
ajoymajumdar: How about using Joda Instant?

metacat-common/src/main/java/com/netflix/metacat/common/dto/notifications/sns/SNSMessageType.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @author tgianos
+ * @since 0.1.47
+ */
+public enum SNSMessageType {
+    TABLE_CREATE,
+    TABLE_DELETE,
+    TABLE_UPDATE,
+    TABLE_PARTITIONS_UPDATE,
+    PARTITION_CREATE,
```
ajoymajumdar: PARTITION_ADD

metacat-common/src/main/java/com/netflix/metacat/common/dto/notifications/sns/SNSMessageType.java
```diff
@@ -0,0 +1,33 @@
...
+ * Enumeration of the various types of SNS events there can be.
+ *
+ * @author tgianos
+ * @since 0.1.47
+ */
+public enum SNSMessageType {
+    TABLE_CREATE,
+    TABLE_DELETE,
+    TABLE_UPDATE,
+    TABLE_PARTITIONS_UPDATE,
```
ajoymajumdar: Do we track partition add and delete at the table level?

metacat-common/src/main/java/com/netflix/metacat/common/dto/notifications/sns/SNSMessageFactory.java
```diff
@@ -0,0 +1,83 @@
...
+     * Convert a JSON String into a message if possible.
+     *
+     * @param json The body of the message to convert back to the original message object from JSON string
+     * @return The message bound back into a POJO
+     * @throws IOException When the input isn't valid JSON
+     */
+    public SNSMessage<?> getMessage(final String json) throws IOException {
+        final JsonNode object = this.mapper.readTree(json);
+        if (object.has(TYPE_FIELD)) {
+            final SNSMessageType messageType = SNSMessageType.valueOf(object.get(TYPE_FIELD).asText());
```
ajoymajumdar: Is there a need for sub classes of SNSMessage?

metacat-common/src/main/java/com/netflix/metacat/common/dto/notifications/sns/SNSMessageType.java
tgianos: Is this a rhetorical question? The event that comes in from internal has a qname and only one qname for the entire payload of partitions so I assume that is the qname of the table they're for... no? If that's not the case we probably need a whole other big feature outside the scope of this PR to fix that before we can send the notification of a table changing. Unless you're saying we need to parse each of the partitions to figure out what table they belong to?

metacat-common/src/main/java/com/netflix/metacat/common/dto/notifications/sns/SNSMessageFactory.java
tgianos: I made the sub classes mostly out of convenience for bindings and usage by clients. It allows them to do instance of and not have complicated payload bindings to deal with as the types are already set. Functionally they're just wrappers as you noticed but just there for convenience.

metacat-main/src/main/java/com/netflix/metacat/main/services/notifications/sns/SNSNotificationsModule.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+@Slf4j
+public class SNSNotificationsModule extends AbstractModule {
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected void configure() {
+        final Multibinder<NotificationService> notificationServices = Multibinder.newSetBinder(
```
ajoymajumdar: Nice.

metacat-main/src/main/java/com/netflix/metacat/main/services/notifications/sns/SNSNotificationsModule.java
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * If SNS is enabled this will return an instance of the SNS Notification service implementation.
+     *
+     * @param config       The metacat configuration
+     * @param objectMapper The JSON object mapper to use
+     * @return If enabled a SNS Notification Service implementation. Otherwise null.
+     */
+    @Provides
+    public NotificationService getSnsNotificationService(
```
ajoymajumdar: Where is this provider used?

tgianos: Whoops it's not. I forgot to delete it after I moved it into its own class. Good catch will remove

PR #19, opened 2016-12-05
elastic search migration util

zhljen:
Adding functions to allow migrating elastic search to an new index.


metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchMigrationUtil.java
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * Constructor.
+     * @param client elastic search client
+     * @param config metacat server config
+     * @param metacatJson json utility
+     */
+    @Inject
+    public ElasticSearchMigrationUtil(
+        @Nullable
+        final Client client,
```
tgianos: Why would the elastic search client be nullable? I would think this would have to be marked `@NotNull` since you can't do anything with elasticsearch without a client

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchMigrationUtil.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.List;
+import java.util.ArrayList;
+
+/**
+ * Class for migrating the elastic search metacat index to a new one. During the migration period, metacat
+ * writes and updates the data to both the existing and the new merge index. This class overrides the
+ * save, update, and softdelete methods. For the save op, it saves to bothe indices. for the update and softdelete,
+ * it first executes the normal index ops, then copies to the merge index
+ */
+public class ElasticSearchMigrationUtil extends ElasticSearchUtil {
```
tgianos: This may be a stupid statement but it seems to me this class really acts more like an aspect where when enabled it needs to perform some added action above and beyond the original action in the base class. That said why don't we just make an aspect where after every call on the core ElasticSearchUtil an aspect is enabled (when the migration is enabled) to invoke the copy from the base index to the new index. Basically almost a retry that does the same actions on the core index with a secondary index. Or just re-write the core elasticsearch util to allow you to send in the write indices and then iterate through them for all the methods in there at runtime. Then you just change the write-indices using a setter when the migration is active and remove the old index from the list once the migration is done. This would save from maintaining two classes and overriding methods, etc.

zhljen: Yes, we have been discussing this. The idea to separate them into two classes is to make the code cleaner. However, I don't see it actually bring much value given the unnecessary complexity bought into the binding phase. I think using one class is better. Also, this migration feature is rarely used. It makes sense to me to limit the 'migration' related code in one implementation class, and the upstream application should not know this.

The way to implement the migration is a bit different from writing to multiple indices using a loop. The reason is that we need to using softdelete and update, which will fail if the data were not there. Thus, for the methods save, we are saving them into the secondary indexes after saving to the original one. For the softdelete and updates, we are making the changes to the existing one, and then copy over to the secondary index. Otherwise, we will need to first copy them and then executing softdelete/update which is inefficient.

zhljen:
Delete the unneeded pull request.

PR #23, opened 2016-12-12
supporing managed table for hive

zhljen:
Changes to fix [LA-1911](https://jira.netflix.com/browse/LA-1911) which allows hive support managed table instead of external table only.


zhljen:
Removing the jira link from the comments.

PR #24, opened 2016-12-12
Minor fixes.

ajoymajumdar:
1. Fixed the log message when loading catalogs.
2. Modified ElasticSearchUtilImpl to import the right Strings class.


zhljen:
Looks good.

PR #29, opened 2017-01-09
Typeconverters

zhljen:
Initial version for canonical type changes. 
- metacat-schema-converters: schema converters between metacat and others, only have hive for now
- metacat-type-converters: type converters between canonical and others
- Removed the presto dependencies on the type converters.


metacat-common/src/main/java/com/netflix/metacat/common/dto/AuditDto.java
```diff
@@ -26,6 +29,9 @@
 @SuppressWarnings("unused")
 @Data
 @EqualsAndHashCode(callSuper = false)
```
tgianos: Why don't we want to call super here? I would think it would be important to know that the base class members are equal as well, no?

metacat-schema-converters/build.gradle
```diff
@@ -0,0 +1,16 @@
...
+
+dependencies {
+    compile project(':metacat-common')
+    compile project(':metacat-type-converters')
+    compile("org.apache.hive:hive-metastore:1.2.1")
+    compile "com.google.guava:guava:${guava_version}"
+    compile "com.google.inject:guice:${guice_version}"
+    compile 'com.netflix.archaius:archaius-core:0.6.5'
+    compile 'com.netflix.servo:servo-core:0.8.3'
+    compile 'org.apache.pig:pig:0.14.0'
```
tgianos: Do we really need to rely on pig? that seems like a very large dependency that we may want to make pluggable in the design... like you can add type converter modules optionally but those modules should pull in these types of dependencies not the core module

metacat-schema-converters/src/main/java/com/netflix/metacat/shema/converters/HiveMetacatConverters.java
```diff
@@ -0,0 +1,90 @@
...
+import org.apache.hadoop.hive.metastore.api.Database;
+import org.apache.hadoop.hive.metastore.api.Partition;
+import org.apache.hadoop.hive.metastore.api.Table;
+
+import java.util.List;
+
+/**
+ * Class to convert hive schema to metacat DTO and vice versa.
+ */
+public interface HiveMetacatConverters {
```
tgianos: Again to me this seems like a module that should be outside the core. We need a very generic interface... @ajoymajumdar might have different opinion but I think all the type converters should implement a common interface and we shouldn't have specific interfaces for hive, pig, etc. just implementations of the common interfaces

metacat-schema-converters/src/main/java/com/netflix/metacat/shema/converters/HiveMetacatConvertersImpl.java
```diff
@@ -0,0 +1,346 @@
...
+
+/**
+ * Class implements hiveMetacatConverter interface.
+ */
+public class HiveMetacatConvertersImpl implements HiveMetacatConverters {
+    private static final Splitter SLASH_SPLITTER = Splitter.on('/');
+    private static final Splitter EQUAL_SPLITTER = Splitter.on('=').limit(2);
+
+    @VisibleForTesting
+    Integer dateToEpochSeconds(final Date date) {
```
tgianos: If this is package private why do you need `@VisableForTesting`? Should be accessible from the test class cause it's in the same package.

metacat-schema-converters/src/main/java/com/netflix/metacat/shema/converters/HiveMetacatConvertersImpl.java
```diff
@@ -0,0 +1,346 @@
...
+
+    private FieldSchema metacatToHiveField(final FieldDto fieldDto) {
+        final FieldSchema result = new FieldSchema();
+        result.setName(fieldDto.getName());
+        result.setType(fieldDto.getType());
+        result.setComment(fieldDto.getComment());
+        return result;
+    }
+
+    @Override
```
tgianos: To these methods you should add javadoc inheritance
```
/**
 * {@inheritdoc}
 *
 */
```

metacat-common/src/main/java/com/netflix/metacat/common/dto/AuditDto.java
zhljen: This is existing code and I will take a further look.

metacat-schema-converters/build.gradle
zhljen: Have removed this module based on discussion.

metacat-schema-converters/src/main/java/com/netflix/metacat/shema/converters/HiveMetacatConverters.java
zhljen: Agreed. Moved it into hive connector.

metacat-schema-converters/src/main/java/com/netflix/metacat/shema/converters/HiveMetacatConvertersImpl.java
zhljen: Agreed.

PR #31, opened 2017-01-18
1. Moved the logic of event notification to the service layer.

ajoymajumdar:
2. Removed the logic to listen to metacat mview events and store it in elastic search. Since these mviews are anyways stored as tables under franklinviews.

### DIFF
--- a/metacat-main/src/main/java/com/netflix/metacat/main/services/CatalogService.java
+++ b/metacat-main/src/main/java/com/netflix/metacat/main/services/CatalogService.java
@@ -31,9 +31,7 @@ public interface CatalogService {
      * @return the information about the given catalog
      */
     @Nonnull
-    CatalogDto get(
-        @Nonnull
-            QualifiedName name);
+    CatalogDto get(@Nonnull QualifiedName name);
 
     /**
      * List of registered catalogs.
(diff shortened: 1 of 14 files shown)


metacat-main/src/main/java/com/netflix/metacat/main/services/MViewService.java
```diff
@@ -35,61 +36,54 @@
...
 
     /**
      * Get the view for the given name.
      * @param name name
      * @return view
      */
-    Optional<TableDto> getOpt(
-        @Nonnull
-            QualifiedName name);
+    Optional<TableDto> getOpt(@Nonnull QualifiedName name);
```
tgianos: why is this `getOpt`? seems like an odd name giving the description says you're getting the view for a name?

metacat-main/src/main/java/com/netflix/metacat/main/services/CatalogService.java
```diff
@@ -31,9 +31,7 @@
      * @return the information about the given catalog
      */
     @Nonnull
-    CatalogDto get(
-        @Nonnull
-            QualifiedName name);
+    CatalogDto get(@Nonnull QualifiedName name);
```
tgianos: Shouldn't we get consistent about using `@Nonnull` vs `@NotNull`

metacat-main/src/main/java/com/netflix/metacat/main/services/MViewService.java
```diff
@@ -102,9 +96,7 @@ void deletePartitions(
      * @param includePartitionDetails if true, includes parameter details
      * @return list of partitions
      */
-    List<PartitionDto> listPartitions(
-        @Nonnull
-            QualifiedName name, String filter, List<String> partitionNames,
+    List<PartitionDto> listPartitions(@Nonnull QualifiedName name, String filter, List<String> partitionNames,
```
tgianos: Condense into the `getPartitionsRequestDto`?

PR #45, opened 2017-02-13
Adding unit tests for connector-hive

zhljen:
Adding unit tests for connector-hive and other minor fixes


metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/model/BaseInfo.java
```diff
@@ -15,20 +15,18 @@
...
 import java.util.Map;
 
 /**
  * Base class for catalog resources.
  * @author amajumdar
  */
-@Getter
-@Setter
-@AllArgsConstructor
+@Data
```
tgianos: Using `@Data` is a bad idea for these type classes as they're meant to transfer data and should be immutable. `@Data` includes setters. All fields should be private final.

metacat-connector-hive/build.gradle
```diff
@@ -13,6 +13,7 @@
 
 dependencies {
     compile project(':metacat-common-server')
+    testCompile project(':metacat-common-server').sourceSets.test.output
```
tgianos: Why are you including this like this?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorDatabaseService.java
```diff
@@ -42,7 +44,11 @@
  * @author zhenl
  */
 public class HiveConnectorDatabaseService implements ConnectorDatabaseService {
+    @Getter
+    @Setter
```
tgianos: Why is there a setter here? On final fields you won't ever be able to change them after they're initially set.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorDatabaseService.java
```diff
@@ -122,7 +128,11 @@ public boolean exists(@Nonnull final ConnectorContext requestContext, @Nonnull f
         try {
             List<QualifiedName> qualifiedNames = Lists.newArrayList();
             for (String databaseName : metacatHiveClient.getAllDatabases()) {
-                qualifiedNames.add(QualifiedName.ofDatabase(name.getCatalogName(), databaseName));
+                final QualifiedName qualifiedName = QualifiedName.ofDatabase(name.getCatalogName(), databaseName);
+                if (prefix != null && !qualifiedName.toString().startsWith(prefix.toString())) {
```
tgianos: This is somewhat innefficiant cause you're basically making a string comparison on a longer string `catalogName/databasePrefix` vs just comparing the prefix against the database name. I'm not sure how we want do do it but rather than doing this comparison over and over I would just once outside the loop compare catalogs and make sure they match and then only compare the actual database names in here.

metacat-connector-hive/src/test/groovy/com/netflix/metacat/connector/hive/HiveConnectorDatabaseSpec.groovy
```diff
@@ -0,0 +1,119 @@
...
+    ConnectorContext connectorContext = new ConnectorContext(1, null);
+
+    def setupSpec() {
+        metacatHiveClient.getAllDatabases() >> ["test1", "test2", "dev1", "dev2"]
+        metacatHiveClient.getDatabase("testdb2") >> { throw new TException() }
+        metacatHiveClient.createDatabase(new Database("testdb3","testdb3","",Collections.EMPTY_MAP)) >> {}
+        metacatHiveClient.createDatabase(new Database("testdb2","testdb2","",Collections.EMPTY_MAP)) >> { throw new TException()}
+    }
+
+    @Unroll
```
tgianos: why is unroll on all of these when you're not actually running multiple tests using a table?

metacat-connector-hive/build.gradle
zhljen: To allow the test to use the static testing data provided in other module. If we have any other way to achieve this, I will make the change to use it.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorDatabaseService.java
zhljen: yes, I agree. I will change this.

PR #57, opened 2017-03-03
Fixing thrift functional test 

zhljen:


metacat-connector-hive/build.gradle
```diff
@@ -13,11 +13,14 @@
 
 dependencies {
     compile project(':metacat-common-server')
-    compile "org.apache.hive:hive-metastore:${hive_version}"
+    compile("org.apache.hive:hive-metastore:${hive_version}") {
+        exclude group: 'com.google.guava'
+    }
+    compile("com.facebook.presto.hive:hive-apache:${facebook_hive_apache}")
```
ajoymajumdar: Why do we need this library?

metacat-connector-hive/build.gradle
```diff
@@ -13,11 +13,14 @@
...
     compile "org.apache.hadoop:hadoop-core:${hadoopcore_version}"
     compile "com.google.guava:guava:${guava_version}"
     compile "com.google.inject:guice:${guice_version}"
     compile "com.google.inject.extensions:guice-persist:${guice_version}"
     compile "com.google.inject.extensions:guice-multibindings:${guice_version}"
-    testCompile project(':metacat-common-server').sourceSets.test.output
+    testCompile project(':metacat-testdata-provider').sourceSets.test.output
```
ajoymajumdar: Nice.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorModule.java
```diff
@@ -55,7 +57,9 @@ public HiveConnectorModule(final String catalogName, final Map<String, String> c
         this.catalogName = catalogName;
         this.infoConverter = infoConverter;
         this.hiveMetastoreClientFactory =
-                new HiveMetastoreClientFactory(null);
+                new HiveMetastoreClientFactory(null,
+                        (int) HiveConnectorUtil.toTime(configuration.getOrDefault(hivemetastoreTimeout, "20s"),
```
ajoymajumdar: Can we consider the configuration as time in seconds. This way you dont have to parse the string for the unit.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorTableService.java
```diff
@@ -108,6 +128,116 @@ public void create(@Nonnull final ConnectorContext requestContext, @Nonnull fina
...
+        final StorageDescriptor sd = table.getSd() != null ? table.getSd() : new StorageDescriptor();
+        String inputFormat = null;
+        String outputFormat = null;
+        Map<String, String> sdParameters = Maps.newHashMap();
+        final String location =
+                tableInfo.getSerde() == null ? null : tableInfo.getSerde().getUri();
+        if (location != null) {
+            sd.setLocation(location);
+        } else if (sd.getLocation() == null) {
+            final String locationStr = hiveConnectorDatabaseService.get(requestContext,
```
ajoymajumdar: I do understand we spoke about this. Cant we use the hive client to get the database?

metacat-connector-hive/build.gradle
zhljen: The HiveStorageFormat.java needs the class in the jar. We can put in the hive-exec jar, which is a fat jar causing dependencies issue.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorModule.java
zhljen: I think this is better as we can tell the unit from the configuration.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorTableService.java
zhljen: Yes, we can, but it will duplicate the code for exception handling.

metacat-connector-hive/build.gradle
ajoymajumdar: I prefer hive-exec jar.hive-apache is a mix of various jars.

zhljen: that's jar bringing in the problem of guava version issue. My guess is that there's a jar having guava inside and the conflicts resolution isn't able to resolve it. We can take a further look if you want.

ajoymajumdar: Even hive-apache depends on hive-exec but excludes certain dependencies. https://github.com/prestodb/presto-hive-apache/blob/master/pom.xml

PR #67, opened 2017-03-15
Cassandra database and table services

tgianos:


metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorUtils.java
```diff
@@ -0,0 +1,96 @@
...
+    public static <T> List<T> paginate(
+        @Nonnull @NonNull final List<T> elements,
+        @Nullable final Pageable pageable
+    ) {
+        final ImmutableList.Builder<T> results = ImmutableList.builder();
+        if (pageable != null && pageable.isPageable()) {
+            results.addAll(
+                elements
+                    .stream()
+                    .skip(pageable.getOffset() * pageable.getLimit())
```
ajoymajumdar: Offset is the row offset and not the page offset

PR #68, opened 2017-03-15
Adding internal hive connector

zhljen:


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorFactory.java
```diff
@@ -95,7 +99,11 @@ private void init() {
     private IMetacatHiveClient createLocalClient() throws MetaException {
         final HiveConf conf = getDefaultConf();
         configuration.forEach(conf::set);
-        return new EmbeddedHiveClient(MetacatHMSHandler.newRetryingHMSHandler("metacat", conf, true));
+        DataSourceManager.get().load(catalogName, configuration);
```
ajoymajumdar: We will have to rethink the use of DataSourceManager. For now, lets add a TODO item.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorDatabaseService.java
```diff
@@ -108,6 +110,24 @@ public void delete(@Nonnull final ConnectorContext requestContext, @Nonnull fina
      * {@inheritDoc}.
      */
     @Override
+    public void update(@Nonnull final ConnectorContext context, @Nonnull final DatabaseInfo databaseInfo) {
```
tgianos: I'd recommend adding Lombok `@NonNull` to these fields so your code is precondition checked to avoid NPE

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/util/HivePartitionName.java
```diff
@@ -0,0 +1,45 @@
+/*
```
ajoymajumdar: Why do we need this class?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorDatabaseService.java
```diff
@@ -108,6 +110,24 @@ public void delete(@Nonnull final ConnectorContext requestContext, @Nonnull fina
...
+        try {
+            this.metacatHiveClient.alterDatabase(databaseInfo.getName().getDatabaseName(),
```
tgianos: You call `getName()` a lot in here. Might be better to save it to a local variable. The compiler probably optimizes this to squash the function call but if we ever change the implementation of `getName` to do some processing under the hood you would replicate executing that functionality multiple times this way. If you put it in a local variable you only have to do it one time. good overall advice

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/util/HiveTableName.java
```diff
@@ -0,0 +1,45 @@
+/*
```
ajoymajumdar: Why do we need this class? We can use QualifiedName.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/util/RetryHelper.java
```diff
@@ -0,0 +1,227 @@
+/*
```
ajoymajumdar: We should use https://github.com/rholder/guava-retrying.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorFactory.java
```diff
@@ -75,17 +79,17 @@ public HiveConnectorFactory(final String name, final Map<String, String> configu
     private void init() {
```
tgianos: I'm not sure I understand the purpose of an `init()` method that is only called from the constructor anyway. Why not just do the logic in the constructor? I've seen @ajoymajumdar do this as well so he might have a reason

ajoymajumdar:
How are we handling sorting and pagination in the embedded hive client?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorFactory.java
```diff
@@ -75,17 +79,17 @@ public HiveConnectorFactory(final String name, final Map<String, String> configu
...
             }
         } catch (Exception e) {
             throw new IllegalArgumentException(
-                String.format("Failed creating the hive metastore client for catalog: %s", name), e);
+                    String.format("Failed creating the hive metastore client for catalog: %s", catalogName), e);
         }
-        final Module hiveModule = new HiveConnectorModule(name, configuration, infoConverter, client);
+        final Module hiveModule = new HiveConnectorModule(catalogName, configuration, infoConverter, client);
         final Injector injector = Guice.createInjector(hiveModule);
         this.databaseService = injector.getInstance(ConnectorDatabaseService.class);
```
tgianos: While this is probably fine cause we know the inner workings of the module/injector and you're just going to return a singleton it's probably a bad practice in general because if the injector really returns a different instance every time with different logic (e.g. a provider method) then users wouldn't get new updated versions of the database service every time. They'd be stuck with the old one.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorFactory.java
```diff
@@ -106,15 +114,17 @@ private static HiveConf getDefaultConf() {
         result.setInt("javax.jdo.option.DatastoreWriteTimeoutMillis", 60000);
```
tgianos: I'd make all these strings constants with names as well as their default values you're setting here. That way we can quickly change them in one place if we have to use them in multiple places.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorFactory.java
```diff
@@ -144,15 +154,15 @@ public ConnectorPartitionService getPartitionService() {
...
+        return catalogName;
     }
 
     @Override
     public void stop() {
         try {
             client.shutdown();
         } catch (TException e) {
-            log.warn(String.format("Failed shutting down the catalog: %s", name), e);
+            log.warn(String.format("Failed shutting down the catalog: %s", catalogName), e);
```
tgianos: You don't have to embed a String.format inside the slf4j logs if you're just doing string replacement. It's redundant. you could change this to be `"Failed shutting down the catalog: {}", catalogName, e` and it would be the same.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/client/embedded/HivePersistenceManagerFactory.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    private static synchronized PersistenceManagerFactory getpersistencemanagerfactory(final Map props) {
+        final String name = String.valueOf(props.get("javax.jdo.option.name"));
+        PersistenceManagerFactory result = factories.get(name);
+        if (result == null) {
+            final DataSource dataSource = DataSourceManager.get().get(name);
+            final String jdoTimeout = JDO_TIMEOUT.get();
+            final Map<String, Object> properties = Maps.newHashMap();
+            properties.put("datanucleus.autoCreateSchema", props.get("datanucleus.autoCreateSchema"));
```
tgianos: These should all probably be constants.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/client/embedded/HivePrivilege.java
```diff
@@ -0,0 +1,66 @@
...
+    /**GRANT.*/
+    GRANT;
+
+    /**
+     * parsePrivilege.
+     * @param userGrant userGrant
+     * @return set of hive privilege
+     */
+    public static Set<HivePrivilege> parsePrivilege(final PrivilegeGrantInfo userGrant) {
+        final String name = userGrant.getPrivilege().toUpperCase(java.util.Locale.ENGLISH);
```
tgianos: I don't quite understand this switch statement. It seems you can either have 1 or all but no other permutations?

PR #71, opened 2017-03-18
using guava retry logic and adding interal sql based hive table parti…

zhljen:


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorFactory.java
```diff
@@ -73,9 +70,13 @@ public HiveConnectorFactory(@Nonnull @NonNull final String catalogName,
         this.infoConverter = infoConverter;
         try {
             final boolean useLocalMetastore = Boolean
-                    .parseBoolean(configuration.getOrDefault(USE_EMBEDDED_METASTORE, "false"));
+                    .parseBoolean(configuration.getOrDefault(HiveConfigConstants.USE_EMBEDDED_METASTORE, "false"));
             if (!useLocalMetastore) {
                 client = createThriftClient();
+                if (configuration.containsKey(HiveConfigConstants.USE_FASTHIVE_SERVICE)) {
```
ajoymajumdar: Does this look right?

ajoymajumdar: Do we need to expose this as a configuration?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/client/embedded/EmbeddedHiveClient.java
```diff
@@ -438,44 +359,75 @@ public int getPartitionCount(@Nonnull @NonNull final String databaseName,
...
+     * Wrapper class for retry execution with exception handling.
+     * @param functionName functionName
+     * @param callable callable
+     * @param classes exception class
+     * @param <V> v
+     * @return function call result, Void in case null
+     * @throws TException
+     * @throws ConnectorException
+     */
+    private <V> V retryExec(final String functionName,
```
ajoymajumdar: We should not be building the retryer for every call.

PR #75, opened 2017-03-22
Add exception mapper to Cassandra Connector

tgianos:


metacat-connector-cassandra/src/main/java/com/netflix/metacat/connector/cassandra/CassandraConnectorDatabaseService.java
```diff
@@ -85,8 +96,13 @@ public void create(
     public void delete(@Nonnull @NonNull final ConnectorContext context, @Nonnull @NonNull final QualifiedName name) {
         final String keyspace = name.getDatabaseName();
         log.debug("Attempting to drop Cassandra keyspace {} for request {}", keyspace, context);
-        this.executeQuery("DROP KEYSPACE IF EXISTS " + keyspace + ";");
-        log.debug("Successfully dropped {} keyspace", keyspace);
+        try {
+            this.executeQuery("DROP KEYSPACE IF EXISTS " + keyspace + ";");
+            log.debug("Successfully dropped {} keyspace", keyspace);
```
zhljen: What if the KEYSPACE does not exist? 'IF EXISTS' will make it as a no-op

metacat-connector-cassandra/src/main/java/com/netflix/metacat/connector/cassandra/CassandraConnectorTableService.java
```diff
@@ -77,8 +80,13 @@ public void delete(@Nonnull @NonNull final ConnectorContext context, @Nonnull @N
         final String keyspace = name.getDatabaseName();
         final String table = name.getTableName();
         log.debug("Attempting to delete Cassandra table {}.{} for request {}", keyspace, table, context);
-        this.executeQuery("USE " + keyspace + "; DROP TABLE IF EXISTS " + table + ";");
-        log.debug("Successfully deleted Cassandra table {}.{} for request {}", keyspace, table, context);
+        try {
+            this.executeQuery("USE " + keyspace + "; DROP TABLE IF EXISTS " + table + ";");
+            log.debug("Successfully deleted Cassandra table {}.{} for request {}", keyspace, table, context);
```
zhljen: Similar as above .

metacat-connector-cassandra/src/main/java/com/netflix/metacat/connector/cassandra/CassandraExceptionMapper.java
```diff
@@ -0,0 +1,64 @@
...
+     * @param de   The Cassandra driver exception
+     * @param name The fully qualified name of the resource which was attempting to be accessed or modified at time of
+     *             error
+     * @return A connector exception wrapping the DriverException
+     */
+    public ConnectorException toConnectorException(
+        @Nonnull @NonNull final DriverException de,
+        @Nonnull @NonNull final QualifiedName name
+    ) {
+        if (de instanceof AlreadyExistsException) {
```
zhljen: What about the TableNotFoundException and DatabaseNotFoundException?

tgianos: Those are done as part of the API's directly due to how the driver API's are written. The driver will return null on requests for Keyspace (database) or tables that don't exist. In this case there is no need to go to the exception mapper because no exception is thrown by the driver.

metacat-connector-cassandra/src/main/java/com/netflix/metacat/connector/cassandra/CassandraConnectorDatabaseService.java
tgianos: Yes. A no-op which results in the users desired end state anyway. That being that the Database (keyspace) doesn't exist. It's cleaner than throwing an exception just to say what they wanted couldn't happen cause they're already in their desired state.

PR #78, opened 2017-03-24
Dependency management

tgianos:
Leverage Spring dependency management BOM's for keeping all dependencies in sync.


ajoymajumdar:
Nice changes. Why are we hard-coding some of the version numbers. I do agree for a few we need to but I am asking for the most common libs.

tgianos:
@ajoymajumdar the ones in the properties block are either there cause they aren't covered by the BOM's currently or because we currently need to override a bunch to back down to compatible versions. Once we upgrade our library stack beyond 1.0.0 release we should be able to parse this list down

PR #84, opened 2017-03-30
Adding logging and metrics for event subscribers

tgianos:
Also shutdown EventBus


ajoymajumdar:
Do we want to add logging and counters on postAsync and post methods of MetacatEventBus

tgianos:
Sure I can do that

PR #87, opened 2017-03-31
adding sort to hive connector table database partition servcies

zhljen:


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorDatabaseService.java
```diff
@@ -188,6 +189,10 @@ public boolean exists(@Nonnull @NonNull final ConnectorContext requestContext,
                 }
                 qualifiedNames.add(qualifiedName);
             }
+            //supporting sort by qualified name only
+            if (sort != null) {
+                ConnectorUtils.sort(qualifiedNames, sort, Comparator.comparing(QualifiedName::toString));
```
tgianos: You want to sort on the whole qualified name here or just the name of the database itself? I ask cause it's less efficient to turn the catalog and database name into a string and then do the comparison rather than just comparing on the database name only

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorPartitionService.java
```diff
@@ -202,6 +202,13 @@ public int getPartitionCount(
                     filteredPartitionList.add(partition);
                 }
             });
+            if (sort != null) {
+                if (sort.getOrder() == SortOrder.DESC) {
+                    Collections.sort(filteredPartitionList, Collections.reverseOrder());
```
tgianos: What is this sorting on? does partition implement a compare to?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorTableService.java
```diff
@@ -325,6 +326,10 @@ public void update(@Nonnull @NonNull final ConnectorContext requestContext,
                     qualifiedNames.add(qualifiedName);
                 }
             }
+            ////supporting sort by qualified name only
+            if (sort != null) {
+                ConnectorUtils.sort(qualifiedNames, sort, Comparator.comparing(QualifiedName::toString));
```
tgianos: Again may be inefficient here.

tgianos:
General comment: when you open PR's please assign at least a milestone to them so when it comes to release time we can point to everything that has happened within a given release.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorPartitionService.java
zhljen: yes, the partition implements the CompareTo

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorTableService.java
zhljen: Both database and table name are part of the sort input.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorPartitionService.java
tgianos: I still think you should use the comparator here cause it'll bring more flexibility if we ever want to implement something else than just relying on whatever the compareTo has. I guess for now though it's a change either way so up to you

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorDatabaseService.java
zhljen: Sure, I use the whole qualified name for sorting, even though the catalog name should be the same for now.

PR #93, opened 2017-04-05
fixing hive table fields from deserializer

zhljen:


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/converters/HiveConnectorInfoConverter.java
```diff
@@ -109,6 +111,17 @@ public Database fromDatabaseInfo(final DatabaseInfo databaseInfo) {
     @Override
     public TableInfo toTableInfo(final QualifiedName name, final Table table) {
         final List<FieldSchema> nonPartitionColumns = table.getSd().getCols();
+        // add the data fields to the nonPartitionColumns
+        if (table.getSd().getColsSize() == 0) {
```
tgianos: is there any chance here of a NPE on getSd()?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/util/HiveTableUtil.java
```diff
@@ -0,0 +1,85 @@
...
+                    "expected STRUCT: %s", inspector.getCategory());
+            return ((StructObjectInspector) inspector).getAllStructFieldRefs();
+        } catch (SerDeException e) {
+            throw Throwables.propagate(e);
+        }
+    }
+
+    private static Class<? extends Deserializer> getDeserializerClass(final String name) {
+        // CDH uses different names for Parquet
+        if ("parquet.hive.serde.ParquetHiveSerDe".equals(name)) {
```
tgianos: The compiler probably does this anyway but I would make this string a constant and then flip it so it's name.equals(CONSTANT)... makes it more readable. Can name be null? if so you'll need a check there or use Lombok `@NonNull`

zhljen: Agreed.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/converters/HiveConnectorInfoConverter.java
zhljen: It shouldn't. I added the check just in case.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/converters/HiveConnectorInfoConverter.java
```diff
@@ -109,6 +111,17 @@ public Database fromDatabaseInfo(final DatabaseInfo databaseInfo) {
     @Override
     public TableInfo toTableInfo(final QualifiedName name, final Table table) {
         final List<FieldSchema> nonPartitionColumns = table.getSd().getCols();
+        // add the data fields to the nonPartitionColumns
+        if (table.getSd() != null && table.getSd().getColsSize() == 0) {
```
ajoymajumdar: Do you need the null check? Also you could use var nonPartitionColumns size.

PR #96, opened 2017-04-12
Connector property override capability for services

tgianos:
Allow users to override the default services for MySql, Postgres, Cassandra and Redshift connectors.

Properties for the connector catalog can be set to a different class that implements the desired interface in order to override the default. The properties are:
`metacat.connector.databaseService.class`
`metacat.connector.tableService.class`
`metacat.connector.partitionService.class`

### DIFF
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorModule.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorModule.java
@@ -1,0 +1,96 @@
+/*
+ *
+ *  Copyright 2017 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorModule.java
```diff
@@ -0,0 +1,77 @@
...
+     * The key which a user can set a value in a catalog to override the default table service class.
+     */
+    private static final String TABLE_SERVICE_CLASS_KEY = "metacat.connector.tableService.class";
+
+    /**
+     * The key which a user can set a value in a catalog to override the default partition service class.
+     */
+    private static final String PARTITION_SERVICE_CLASS_KEY = "metacat.connector.partitionService.class";
+
+    protected <S extends ConnectorBaseService> Class<? extends S> getServiceClass(
```
ajoymajumdar: Why dont you want to have separate methods?

PR #99, opened 2017-04-14
Modified to delete the dataMetadata delete markers if the markers are false.

ajoymajumdar:
Modified to delete the dataMetadata delete markers if the markers are false.(meaning there exists tables/partitions referring to the uri)


metacat-common-server/src/main/java/com/netflix/metacat/common/usermetadata/UserMetadataService.java
```diff
@@ -42,6 +42,14 @@ void deleteDataMetadatas(
             List<String> uris);
 
     /**
+     * Delete the delete markers for data metadata for the given uris.
+     * @param uris list of uris.
+     */
+    void deleteDataMetadataDeletes(
```
tgianos: this is really strange formatting

PR #101, opened 2017-04-14
fixing hive connector HMSHandler timeout issue, and use add_drop for …

zhljen:
…deleting partitions


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/metastore/HMSHandlerProxy.java
```diff
@@ -0,0 +1,79 @@
...
+    public Object invoke(final Object proxy, final Method method, final Object[] args) throws Throwable {
+        Deadline.registerIfNot(timeout);
+        try {
+            Deadline.startTimer(method.getName());
+            final Object object = method.invoke(metacatHMSHandler, args);
+            Deadline.stopTimer();
+            return object;
+
+        } catch (Exception e) {
+            throw e.getCause();
```
ajoymajumdar: Why are we propagating the error cause? Does the caller get the proxy exception or the real exception from the HMSClient?

zhljen:
The reflection call wraps the real exception, changed to
''InvocationTargetException
to be more explicit.

PR #107, opened 2017-04-27
Adding searchable definition metadata

zhljen:


metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchDoc.java
```diff
@@ -35,12 +38,34 @@
     private String user;
     private boolean deleted;
     private String refreshMarker;
+    private boolean addSearchableDefinitionMetadata;
+
+    /**
+     * Definition Metadata pull out fields.
+     */
+    private String[] definitionMetadataFields = {
```
tgianos: is this a constant? should it be static and final?

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchDoc.java
```diff
@@ -94,6 +120,10 @@ public static ElasticSearchDoc parse(final GetResponse response) {
 
     private ObjectNode toJsonObject() {
         final ObjectNode oMetadata = MetacatJsonLocator.INSTANCE.toJsonObject(dto);
+        if (addSearchableDefinitionMetadata) {
```
tgianos: it's kind of confusing here to have the variable and the function have exactly the same name

PR #109, opened 2017-05-02
Removed the automatic registration of listeners to Metacat event bus.

ajoymajumdar:
Removed the automatic registration of listeners to Metacat event bus. Added more context info to logs and metrics on SNS publish failures.

### DIFF
--- a/metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
+++ b/metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
@@ -42,6 +42,7 @@ public final class QualifiedName implements Serializable {
 
     private String qualifiedName;
     private Map<String, String> qualifiedNameMap;
+    private Map<String, String> parts;
     private Type type;
 
     private QualifiedName(@Nonnull final String catalogName,
@@ -349,7 +350,7 @@ public final class QualifiedName implements Serializable {
     @JsonValue
     public Map<String, String> toJson() {
         if (qualifiedNameMap == null) {
-            final Map<String, String> map = new HashMap<>(4);
+            final Map<String, String> map = new HashMap<>(5);
             map.put("qualifiedName", toString());
             map.put("catalogName", catalogName);
 
@@ -373,6 +374,37 @@ public final class QualifiedName implements Serializable {
         }
 
         return qualifiedNameMap;
+    }
+
+    /**
+     * Returns the qualified name in parts.
+     * @return parts of the qualified name as a Map
+     */
+    public Map<String, String> parts() {
+        if (parts == null) {
+            final Map<String, String> map = new HashMap<>(4);
+            map.put("catalogName", catalogName);
+
+            if (!databaseName.isEmpty()) {
+                map.put("databaseName", databaseName);
+            }
+
+            if (!tableName.isEmpty()) {
+                map.put("tableName", tableName);
+            }
+
+            if (!partitionName.isEmpty()) {
+                map.put("partitionName", partitionName);
+            }
+
+            if (!viewName.isEmpty()) {
+                map.put("viewName", viewName);
+            }
+
+            parts = map;
+        }
+
+        return parts;
     }
 
     public boolean isViewDefinition() {
--- a/metacat-main/src/main/java/com/netflix/metacat/main/services/notifications/sns/SNSNotificationServiceImpl.java
+++ b/metacat-main/src/main/java/com/netflix/metacat/main/services/notifications/sns/SNSNotificationServiceImpl.java
@@ -23,6 +23,8 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.github.fge.jsonpatch.JsonPatch;
 import com.github.fge.jsonpatch.diff.JsonDiff;
+import com.google.common.collect.Maps;
+import com.netflix.metacat.common.QualifiedName;
 import com.netflix.metacat.common.dto.PartitionDto;
 import com.netflix.metacat.common.dto.TableDto;
 import com.netflix.metacat.common.dto.notifications.sns.SNSMessage;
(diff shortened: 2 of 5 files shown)


metacat-main/src/main/java/com/netflix/metacat/main/services/notifications/sns/SNSNotificationServiceImpl.java
```diff
@@ -90,12 +95,13 @@ public SNSNotificationServiceImpl(
     public void notifyOfPartitionAddition(@Nonnull final MetacatSaveTablePartitionPostEvent event) {
         log.debug("Received SaveTablePartitionPostEvent {}", event);
         CounterWrapper.incrementCounter("metacat.notifications.sns.events.partitions.add");
+        final String name = event.getName().toString();
+        final long timestamp = event.getRequestContext().getTimestamp();
+        final String requestId = event.getRequestContext().getId();
+        AddPartitionMessage message = null;
```
tgianos: You should be able to mark this final and assigning it to null is redundant.

metacat-main/src/main/java/com/netflix/metacat/main/services/notifications/sns/SNSNotificationServiceImpl.java
```diff
@@ -106,9 +112,14 @@ public void notifyOfPartitionAddition(@Nonnull final MetacatSaveTablePartitionPo
                 log.debug("Published create partition message {} on {}", message, this.partitionTopicArn);
                 CounterWrapper.incrementCounter("metacat.notifications.sns.partitions.add.succeeded");
             }
-
+        } catch (final Exception e) {
+            handleException(event.getName(), "Unable to publish partition creation notification",
+                "metacat.notifications.sns.partitions.add.failed", message, e);
```
tgianos: We should make these strings constants

metacat-main/src/main/java/com/netflix/metacat/main/services/notifications/sns/SNSNotificationServiceImpl.java
```diff
@@ -260,11 +281,23 @@ public void notifyOfTableUpdate(@Nonnull final MetacatUpdateTablePostEvent event
...
         }
     }
 
+    private void handleException(final QualifiedName name, final String message, final String counterKey,
+        final SNSMessage payload, final Exception e) {
+        log.error(String.format("%s with payload: %s", message, payload), e);
+        DynamicCounter.increment(counterKey, getQualifiedNameTagList(name));
+    }
+
+    private static BasicTagList getQualifiedNameTagList(final QualifiedName name) {
```
tgianos: This seems like a utility method that should be more global no? Also why is it static?

metacat-main/src/main/java/com/netflix/metacat/main/services/notifications/sns/SNSNotificationServiceImpl.java
```diff
@@ -260,11 +281,23 @@ public void notifyOfTableUpdate(@Nonnull final MetacatUpdateTablePostEvent event
...
-            log.error("Unable to publish update table notification", e);
-            CounterWrapper.incrementCounter("metacat.notifications.sns.tables.update.failed");
+            handleException(event.getName(), "Unable to publish update table notification",
+                "metacat.notifications.sns.tables.update.failed", message, e);
         }
     }
 
+    private void handleException(final QualifiedName name, final String message, final String counterKey,
+        final SNSMessage payload, final Exception e) {
+        log.error(String.format("%s with payload: %s", message, payload), e);
```
tgianos: Why call String.format. SLF4J will automatically do parameter replacement with `{}` in order. This seems like an unnecessary method call.

metacat-main/src/main/java/com/netflix/metacat/main/services/notifications/sns/SNSNotificationServiceImpl.java
ajoymajumdar: We need to initialize it because we are using it in the exception block.

metacat-main/src/main/java/com/netflix/metacat/main/services/notifications/sns/SNSNotificationServiceImpl.java
ajoymajumdar: I do not see a method with format and throwable as parameters.

tgianos: https://www.slf4j.org/faq.html#paramException

metacat-main/src/main/java/com/netflix/metacat/main/services/notifications/sns/SNSNotificationServiceImpl.java
tgianos: Yeah but in java objects are null by default no?

metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
```diff
@@ -375,6 +376,37 @@ private String standardizeRequired(final String name, final String value) {
...
+    /**
+     * Returns the qualified name in parts.
+     * @return parts of the qualified name as a Map
+     */
+    public Map<String, String> parts() {
+        if (parts == null) {
+            final Map<String, String> map = new HashMap<>(4);
+            map.put("catalogName", catalogName);
+
+            if (!databaseName.isEmpty()) {
```
tgianos: Isn't it possible for this to throw NPE? Perhaps use `StringUtils.isNotBlank`

ajoymajumdar: No. For some reason we initialize these to empty string.

tgianos: Weird. Ok sounds like a bigger refactoring for later.

PR #112, opened 2017-05-08
Beginning of port to Spring Boot

tgianos:
This is just the beginning of work to clean up the codebase and move it to Spring Boot as the foundational platform.

The goal is to make the project simpler and more standard.

This PR will compile however it has many tests marked as `@Ignore` and ALL the functional tests disabled via commenting out the logic of the gradle task.

The intention of this PR is to allow others to begin work on this large task in parallel not to be a fully functional commit.


zhljen:
A good starting point to convert Metacat to Spring-boot.

PR #113, opened 2017-05-10
changing servo to spectator

zhljen:


metacat-common-server/src/main/java/com/netflix/metacat/common/server/monitoring/CounterWrapper.java
```diff
@@ -32,21 +31,23 @@
...
  *
  * @author amajumdar
+ * @since 1.0.0
  */
 @Slf4j
 public final class CounterWrapper {
 
-    private static final LoadingCache<String, Counter> COUNTERS = CacheBuilder.newBuilder()
-        .build(
+    private static final LoadingCache<String, Counter> COUNTER_LOADING_CACHE
```
tgianos: We shouldn't need this class anymore. The Registry already maintains the cache for us.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/monitoring/CounterWrapper.java
```diff
@@ -32,21 +31,23 @@
...
+            = CacheBuilder.newBuilder().build(
             new CacheLoader<String, Counter>() {
                 public Counter load(
-                    @Nonnull final String counterName) {
-                    final Counter counter = Monitors.newCounter(counterName);
-                    DefaultMonitorRegistry.getInstance().register(counter);
-                    return counter;
+                        @Nonnull final String counterName) {
+                    return Spectator.globalRegistry().counter(counterName);
```
tgianos: Please no static references. Impossible to test.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/monitoring/DynamicGauges.java
```diff
@@ -0,0 +1,81 @@
+package com.netflix.metacat.common.server.monitoring;
```
tgianos: Shouldn't need this class.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/util/PoolStatsInterceptor.java
```diff
@@ -38,9 +38,11 @@
     private String metricNameTotal;
     private String metricNameActive;
     private String metricNameIdle;
+    private final DynamicGauges dynamicGauges = new DynamicGauges();
```
tgianos: Should be injecting `Registry` into classes that need it.

metacat-main/src/main/java/com/netflix/metacat/main/api/RequestWrapper.java
```diff
@@ -61,20 +85,23 @@ public static QualifiedName qualifyName(final Supplier<QualifiedName> nameSuppli
...
-        DynamicCounter.increment("dse.metacat.counter.requests", tags);
-        final Stopwatch timer = DynamicTimer.start("dse.metacat.timer.requests", tags);
+    public <R> R processRequest(
+            final QualifiedName name,
+            final String resourceRequestName,
+            final Supplier<R> supplier) {
+        final long start = System.currentTimeMillis();
+        final Map<String, String> tags = new HashMap<String, String>(name.parts());
+        tags.put("request", resourceRequestName);
+        registry.counter(requestCounterId.withTags(tags)).increment();
```
tgianos: this is good.

metacat-main/src/main/java/com/netflix/metacat/main/api/RequestWrapper.java
```diff
@@ -87,47 +114,51 @@ public static QualifiedName qualifyName(final Supplier<QualifiedName> nameSuppli
...
+                    e.getCause() == null ? "" : e.getCause().getMessage(), resourceRequestName, name);
             log.error(message, e);
             throw new MetacatException(message, Response.Status.INTERNAL_SERVER_ERROR, e);
         } finally {
-            timer.stop();
+            final long duration = System.currentTimeMillis() - start;
             log.info("### Time taken to complete {} is {} ms", resourceRequestName,
-                timer.getDuration(TimeUnit.MILLISECONDS));
+                    duration);
+            this.registry.timer(requestTimerId).record(duration, TimeUnit.MILLISECONDS);
```
tgianos: good.

metacat-main/src/main/java/com/netflix/metacat/main/api/RequestWrapper.java
```diff
@@ -27,31 +29,53 @@
...
+
+    /**
+     * Wrapper class for processing the request.
+     *
+     * @param registry registry
+     */
+    @Inject
+    public RequestWrapper(@NotNull @NonNull final Registry registry) {
+        this.registry = registry;
+        requestCounterId = registry.createId("dse.metacat.counter.requests");
```
tgianos: I think we should take the chance now to change these metrics names from `dse.*` to just `metacat.*`

metacat-common-server/src/main/java/com/netflix/metacat/common/server/util/PoolStatsInterceptor.java
```diff
@@ -58,9 +61,9 @@ public void disconnected(final ConnectionPool parent, final PooledConnection con
 
     private void publishMetric(final ConnectionPool parent) {
         if (parent != null && metricNameTotal != null) {
-            DynamicGauge.set(metricNameTotal, parent.getSize());
-            DynamicGauge.set(metricNameActive, parent.getActive());
-            DynamicGauge.set(metricNameIdle, parent.getIdle());
+            registry.gauge(registry.createId(metricNameTotal), parent.getSize());
```
ajoymajumdar: Can we create the Id once?

brharrington: Spectator supports an active and passive gauge. Active gauges are similar to counters and other types where the user calls set when something happens (`publishMetric` in this case). The passive case you provide the number reference or a function to compute the number and spectator will poll it in the background until that reference is garbage collected. Based on the usage pattern of the previous you would want the activity based notion here. So it would look like:

```java
registry.gauge(registry.createId(metricNameTotal)).set(parent.getSize());
```

It is unfortunate they appear so similar in the api, but it has to do with how spectator evolved and there are some backwards compatibility constraints at this point.

You can cache the id, but in this case since the id doesn't need to be updated it would be better to just cache the gauge reference. Something like:

```java
private Gauge totalGauge;
...

private void publishMetric(final ConnectionPool parent) {
  ...
  totalGauge.set(parent.getSize());
  ...
}

public void setMetricName(final String metricName) {
  this.totalGauge = registry.gauge(registry.createId(LogConstants.GaugeConnectionsTotal + "." + metricName));
  ...
}
```

zhljen: Thanks for your comments.

zhljen:
Cancelled this PR.

PR #115, opened 2017-05-17
Spring Boot Configuration

tgianos:
Initial commit of Spring Boot port.

Code compiles and runs via `java -jar`. Many tests are disabled currently and will need to be re-enabled, modified etc. as many of these tests depend on Guice.


build.gradle
```diff
@@ -26,14 +26,14 @@ buildscript {
 
     dependencies {
         classpath("org.springframework.boot:spring-boot-gradle-plugin:${spring_boot_version}")
-        classpath("io.spring.gradle:dependency-management-plugin:1.0.0.RELEASE")
         classpath("org.asciidoctor:asciidoctor-gradle-plugin:1.5.3")
         classpath("gradle.plugin.com.gorylenko.gradle-git-properties:gradle-git-properties:1.4.17")
         classpath("com.netflix.nebula:gradle-aggregate-javadocs-plugin:3.0.1")
     }
```
tgianos: Probably could be... feel free :-)

PR #117, opened 2017-05-18
Re-enabling tests with Spring Boot

tgianos:


metacat-connector-hive/src/test/groovy/com/netflix/metacat/connector/hive/HiveConnectorDatabaseSpec.groovy
```diff
@@ -41,160 +40,174 @@ import spock.lang.Unroll
...
-        new MetaException()          |InvalidMetaException
+        exception           | result
+        new TException()    | ConnectorException
+        new MetaException() | InvalidMetaException
 
     }
 
-    @Ignore
-    def "Test for get database" (){
+    def "Test for get database not found"() {
```
tgianos: @ajoymajumdar @zhljen Not this test and the next "Test for get database" the previous behavior didn't seem to make sense to me. Basically the MetacatHiveClient would return a null for the get database and we'd just return a database info object that regurgitated the QualifiedName from the info converter even though I would assume a null from the MetacatHiveClient should mean the database doesn't actually exist. Hence when I re-enabled this test it kept throwing NullPointerException after I re-enabled it cause I'd added `@NonNull` to the database parameter. If this was some sort of expected behavior we can change it back but like I said it didn't seem to make sense to me.

metacat-main/src/test/groovy/com/netflix/metacat/main/search/BaseEsSpec.groovy
```diff
@@ -24,13 +24,15 @@ import org.elasticsearch.action.admin.indices.exists.indices.IndicesExistsReques
...
 import org.elasticsearch.common.settings.Settings
+import org.junit.Ignore
 import spock.lang.Shared
 import spock.lang.Specification
 
 /**
  * Created by amajumdar on 8/17/15.
  */
-class BaseEsSpec extends Specification {
+@Ignore
```
tgianos: @ajoymajumdar note this is the same change you made in #116 I had to put it in here as I couldn't get the build to pass this step due to the multiple runs issue you were talking about.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorPartitionService.java
```diff
@@ -326,7 +326,7 @@ public PartitionsSaveResponse savePartitions(
             return result;
         } catch (NoSuchObjectException exception) {
             if (exception.getMessage() != null && exception.getMessage().startsWith("Partition doesn't exist")) {
-                throw new PartitionNotFoundException(tableName, null, exception);
+                throw new PartitionNotFoundException(tableName, "", exception);
```
tgianos: @zhljen Note this change as this PartitionNotFoundException expects a partition name field but we were passing null.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorDatabaseService.java
```diff
@@ -121,26 +121,32 @@ public void update(@Nonnull @NonNull final ConnectorContext context,
...
-    public DatabaseInfo get(@Nonnull @NonNull final ConnectorContext requestContext,
-                            @Nonnull @NonNull final QualifiedName name) {
+    public DatabaseInfo get(
+        @Nonnull @NonNull final ConnectorContext requestContext,
+        @Nonnull @NonNull final QualifiedName name
+    ) {
         try {
             final Database database = metacatHiveClient.getDatabase(name.getDatabaseName());
-            return hiveMetacatConverters.toDatabaseInfo(name, database);
+            if (database != null) {
```
tgianos: @ajoymajumdar @zhljen This is the associated null check and exception throw change from my other comment for the test behavior change.

zhljen: If the database does not exists, the hive client will return NoSuchObjectException which will return as DatabaseNotFoundException. Don't think the null check is needed here.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorPartitionService.java
zhljen: Ya, good catch. Passing "" looks right to me.

metacat-connector-hive/src/test/groovy/com/netflix/metacat/connector/hive/HiveConnectorDatabaseSpec.groovy
zhljen: I can't think of any case where the getdatabase returns null. The case with a null as the returned result isn't right. We should remove this case. @ajoymajumdar If I miss anything here, please point out.

tgianos: Your test previously assumed it returned null and defaulted into the code base to use some sort of weird default behavior hence my question

zhljen: Yes, it looks that the code assumes the null database as a valid case and even have a test case to check it. It doesn't seem to be right, and we should remove it.

PR #118, opened 2017-05-19
Changing to use spectator for metrics

zhljen:


metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
```diff
@@ -60,7 +65,7 @@ public MetacatEventBus(
      */
     public void postAsync(final ApplicationEvent event) {
         log.debug("Received request to post an event {} asynchronously", event);
-        CounterWrapper.incrementCounter("metacat.events.async");
+        registry.counter(LogConstants.CounterEventAsync.name()).increment();
```
tgianos: Shouldn't it be called `MetricConstants` not `LogConstants`

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
```diff
@@ -71,7 +76,7 @@ public void postAsync(final ApplicationEvent event) {
      */
     public void postSync(final ApplicationEvent event) {
         log.debug("Received request to post an event {} synchronously", event);
-        CounterWrapper.incrementCounter("metacat.events.sync");
+        registry.counter(LogConstants.CounterEventSync.name()).increment();
```
tgianos: Is it better to do this than to just have a counter reference saved in the constructor as a private member so we don't have to get the name every time? Then we just call increment on the local reference

metacat-common-server/src/main/java/com/netflix/metacat/common/server/monitoring/LogConstants.java
```diff
@@ -13,95 +13,109 @@
...
     /**
      * General logging constants.
      */
-    GlobalPrefix("dse"), AppPrefix(GlobalPrefix + ".metacat"),
+    AppPrefix("metacat"),
+
+    /**
+     * evnets.
+     */
+    CounterEventAsync(AppPrefix + ".events.count.Async"),
```
tgianos: it seems to me that Async vs. Sync could be a tag rather than part of the name.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/monitoring/LogConstants.java
```diff
@@ -13,95 +13,109 @@
...
 
+import com.google.common.collect.ImmutableMap;
+
+import java.util.Map;
+
+//CHECKSTYLE:OFF
 /**
  * Log constants.
  */
 public enum LogConstants {
```
tgianos: Rename to MetricsConstants and why is it an enumeration instead of just a class of string constants?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/monitoring/LogConstants.java
```diff
@@ -13,95 +13,109 @@
...
+    /**
+     * metacat request.
+     */
+    CounterRequestCount(AppPrefix + ".count.request"),
+
+
+    /**
+     * Notifications.
+     */
+    CounterSNSNotificationPartitionAdd(AppPrefix + ".notifications.count.partitionsAdd"),
```
tgianos: I might move count to the end. like partitionsAdd.count

metacat-common-server/src/main/java/com/netflix/metacat/common/server/monitoring/LogConstants.java
```diff
@@ -13,95 +13,109 @@
...
+    CounterElasticSearchTableDelete(AppPrefix + ".elasticsearch.count.tableDelete"),
+    CounterElasticSearchTableSave(AppPrefix + ".elasticsearch.count.tableSave"),
+    CounterElasticSearchTableRename(AppPrefix + ".elasticsearch.count.tableRename"),
+    CounterElasticSearchTableUpdate(AppPrefix + ".elasticsearch.count.tableUpdate"),
+    CounterElasticSearchPartitionSave(AppPrefix + ".elasticsearch.count.partitionSave"),
+    CounterElasticSearchPartitionDelete(AppPrefix + ".elasticsearch.count.partitionDelete"),
+    CounterElasticSearchDelete(AppPrefix + ".elasticsearch.count.esDelete"),
+    CounterElasticSearchBulkDelete(AppPrefix + ".elasticsearch.count.esBulkDelete"),
+    CounterElasticSearchUpdate(AppPrefix + ".elasticsearch.count.esUpdate"),
+    CounterElasticSearchBulkUpdate(AppPrefix + ".elasticsearch.count.esBulkUpdate"),
```
tgianos: isn't having it `elasticsearch` and then `es` prefix on the name itself kind of redundant?

gradle.properties
```diff
@@ -37,4 +37,5 @@ hive_version=1.2.1
 
 ## speed up the build process
 #org.gradle.parallel=true
-org.gradle.jvmargs=-Xmx2G
+org.gradle.jvmargs=-Xmx4G
```
ajoymajumdar: Cant we pass this as command arguments?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/util/PoolStatsInterceptor.java
```diff
@@ -57,10 +61,13 @@ public void disconnected(final ConnectionPool parent, final PooledConnection con
     }
 
     private void publishMetric(final ConnectionPool parent) {
-        if (parent != null && metricNameTotal != null) {
-            DynamicGauge.set(metricNameTotal, parent.getSize());
-            DynamicGauge.set(metricNameActive, parent.getActive());
-            DynamicGauge.set(metricNameIdle, parent.getIdle());
+        if (parent != null
+            && metricNameTotalGauage != null
```
tgianos: Rather than doing all these null checks why wouldn't you just instantiate the gauges once in the constructor and make them final member variables?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorFactory.java
```diff
@@ -90,10 +95,20 @@ public HiveConnectorFactory(
...
-        final Module hiveModule = new HiveConnectorModule(config, catalogName, configuration, infoConverter, client);
+        final Module hiveModule = new HiveConnectorModule(config,
+            catalogName, configuration, infoConverter, client, registry);
         this.injector = Guice.createInjector(hiveModule);
     }
 
+    private IMetacatHiveClient createLocalClient() throws Exception {
+        final HiveConf conf = getDefaultConf();
+        configuration.forEach(conf::set);
+        //TO DO Change the usage of DataSourceManager later
```
tgianos: really minor but you should make this TODO one word so we can find it quickly and IDE's will pick it up

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorFastPartitionService.java
```diff
@@ -180,8 +184,10 @@ public int getPartitionCount(
                 partitionsRequest.getPageable(),
                 partitionsRequest.getIncludePartitionDetails());
         } finally {
+            final long duration = registry.clock().monotonicTime() - start;
             log.info("###### Time taken to complete "
-                + "HiveConnectorFastPartitionService.getPartitions is {} ms", timer.stop());
+                + "HiveConnectorFastPartitionService.getPartitions is {} ms", duration);
+            this.registry.timer(LogConstants.TimerHiveGetPartitions.name()).record(duration, TimeUnit.MILLISECONDS);
```
tgianos: keep reference to timer

metacat-thrift/src/main/java/com/netflix/metacat/thrift/CatalogThriftHiveMetastore.java
```diff
@@ -1704,14 +1709,18 @@ private TException unimplemented(final String methodName, final Object[] args) t
...
             log.error(message, e);
             final MetaException me = new MetaException(message);
             me.initCause(e);
             throw me;
         } finally {
-            log.info("+++ Thrift({}): Time taken to complete {} is {} ms", catalogName, methodName, timer.stop());
+            final long duration = registry.clock().monotonicTime() - start;
+             this.registry.timer(LogConstants.TimerThriftRequest.name()
+                  + "." + methodName).record(duration, TimeUnit.MILLISECONDS);
+            log.info("+++ Thrift({}): Time taken to complete {} is {} ms", catalogName, methodName, duration);
```
tgianos: Do we really want all this to be info level? Seems to me they might be better as debug? Not sure how @ajoymajumdar has this overall

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
zhljen: Agreed.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
zhljen: The spectator itself has the counter reference cached. The counter only works in active mode and introducing a local id variable will not save. The gauge is the only one that has passive mode, where caches the reference at application will be useful.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/monitoring/LogConstants.java
zhljen: It can be defined as a constant class. But it also has methods, and needs to be a singleton. Enum seems ok. The naming is bit misleading, and should be renamed as Metrics

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorFastPartitionService.java
zhljen: Similar as the counter explanation.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/monitoring/LogConstants.java
zhljen: I think it's ok to let them be a separated counter for consistency. There're other cases like this, e.g. tableCreate, tableDelete, etc. A couple of cased the tags used are (1) the status of operations, failure/succeed; (2) the catalog/database/table names in case of the same operation.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/monitoring/LogConstants.java
zhljen: This is for the consistency of naming, i.e. the component.spectatorType.measureName. The type can be count, gauge, timer.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/monitoring/LogConstants.java
zhljen: This is to separated the operations from metacat e.g. database/table or elasticSearch itself.  The elasticsearch prefix indicates the components, and the 'es' indicates the operation target.

metacat-thrift/src/main/java/com/netflix/metacat/thrift/CatalogThriftHiveMetastore.java
ajoymajumdar: Info is fine. All request methods log info messages at the start and end of the request.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/util/PoolStatsInterceptor.java
ajoymajumdar: We cannot since Tomcat pool implementation needs a default empty constructor.

zhljen: They can't be instantiated in the constructor, as the "metricName" are unknown at that time.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorFactory.java
zhljen: Agreed.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/monitoring/LogConstants.java
tgianos: I think we should flip it so it's component.measureName.type... I say this cause when we type it into the atlas search filter later on we can better filter down by the metrics and we'll then see that metric name will have a counter, gauge or timer associated with it as opposed to having to back out from the metric to see other metrics cause they're grouped under type... @ajoymajumdar opinion?

PR #119, opened 2017-05-22
Refactored code in TableServiceImpl to remove the cyclic dependency between TableServiceImpl and MViewServiceImpl.

ajoymajumdar:
1. Delete the views of a table when a table gets deleted.
2. Rename the views of a table when a table gets renamed.


metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatDeleteTablePostEvent.java
```diff
@@ -51,7 +52,27 @@ public MetacatDeleteTablePostEvent(
...
+     * Constructor.
+     *
+     * @param name           name
+     * @param requestContext context
+     * @param source         The source object which threw this event
+     * @param table          table info
+     * @param isMView        true, if the table is a materialized view
+     */
+    public MetacatDeleteTablePostEvent(
+        @Nonnull @NonNull final QualifiedName name,
```
tgianos: I thought we decided not to do this with both the annotations? Don't we just need to add the annotation on the `package-info.java` and leave these annotations out completely unless we're putting `@Nullable` on

ajoymajumdar: We can have another PR to remove all the nonnulls.

PR #121, opened 2017-05-23
Metrics: Renaming measurement and adding more hive metrics

zhljen:
Renaming measurement and adding more hive metrics
Removing unused file ElasticSearchClientProvider.java


metacat-common-server/src/main/java/com/netflix/metacat/common/server/monitoring/Metrics.java
```diff
@@ -28,89 +28,100 @@
...
-    CounterThrift(AppPrefix + ".thrift.count.request"),
+    Thrift("thrift"),
+    CounterThrift(Name(Thrift, Count,   "request")),
 
     /**
      * hive sql lock error.
      */
-    CounterHiveSqlLockError(AppPrefix + ".count.hiveSqlLockError"),
-    CounterHiveExperimentGetTablePartitionsFailure(AppPrefix + ".hive.count.experimentGetPartitionsFailure"),
+    Hive("hive"),
```
tgianos: I think that Hive metrics should be kept within the Hive connector module. They have nothing to do with the main metacat system.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/properties/ArchaiusConfigImpl.java
```diff
@@ -20,6 +20,7 @@
 //    private final DynamicStringProperty defaultTypeConverter;
 //    private final DynamicBooleanProperty isElasticSearchEnabled;
 //    private final DynamicStringProperty elasticSearchIndexName;
+    /* THIS IS USED IN TEST*/
```
tgianos: We should just delete this file.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorFastPartitionService.java
```diff
@@ -146,6 +150,9 @@ public int getPartitionCount(
         @Nonnull @NonNull final ConnectorContext requestContext,
         @Nonnull @NonNull final QualifiedName tableName
     ) {
+        final long start = registry.clock().monotonicTime();
+        final Map<String, String> tags = new HashMap<String, String>();
+        tags.put("request", "getPartitionCount");
```
tgianos: Potential constants?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorFastTableService.java
```diff
@@ -121,6 +167,10 @@ public HiveConnectorFastTableService(
                 .query(conn, queryBuilder.toString(), handler, params.toArray());
         } catch (SQLException e) {
             throw Throwables.propagate(e);
+        } finally {
+            final long duration = registry.clock().monotonicTime() - start;
+            log.debug("### Time taken to complete getTableNames is {} ms", duration);
+            this.registry.timer(requestTimerId.withTags(tags)).record(duration, TimeUnit.MILLISECONDS);
```
tgianos: It would be nice if we could find a way to use Aspects to do this repetitive action over and over again

metacat-common-server/src/main/java/com/netflix/metacat/common/server/monitoring/Metrics.java
zhljen: Sounds reasonable to push each metrics into the individual component. However, it's easy to manage them in a centralized location and reuse the same util functions. The metric class is a enum, which can't be extended. Separation the hive out will result in duplicated util functions.  I think it's ok to define them together.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/properties/ArchaiusConfigImpl.java
zhljen: Sure. Just want to leave it there,  in case to address the dynamic properties before removing the file.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorFastPartitionService.java
zhljen: I feel the same thing. Exploring if any better way to do this.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/monitoring/Metrics.java
tgianos: I disagree. Hive connector is a plugin/spi implementation and shouldn't be leaking into the main code base. If we're keeping it here just for enum convenience then that seems to be a bad design and this should be class that can be extended. @ajoymajumdar have spoken before about the need to remove this kind of leakage.

zhljen: sure. I will move it out to hive :)

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorFastTableService.java
zhljen: I investigated the possibility of using aspects here. The AspectJ can provide the functionality to allow us execute the monitoring code before and after the function call. To get there, we can define the monitor as an aspect, implementing the "advice" code there and defines the joinPoint at each hive function call.  This [article](http://www.concretepage.com/spring/spring-aop-aspectj-annotation-example-with-aspect-pointcut-before-after-around-afterreturning-afterthrowing-advice) has a nice example. 
Spring supports AOP nicely, and most AOP usage is based on Spring. Think we should do it too. 
However, the hive connector is bootstrapped by Guice not by Spring. We need to use the raw AspectJ to write the aspect related codes, which seems a bit painful. At this point, I don't see  much value to use the AspectJ, and converted the hive function names to constants. We can revisit this at some point, and open for discussion.

tgianos: Yeah I think we should definitely use it in the metacat-main services or API tier. Can be a separate thing. We can file issue.

PR #123, opened 2017-05-25
making the user metadata service optional

zhljen:


metacat-common-server/src/main/java/com/netflix/metacat/common/server/usermetadata/DefaultUserMetadataService.java
```diff
@@ -0,0 +1,23 @@
...
+ *          limitations under the License.
+ */
+
+package com.netflix.metacat.common.server.usermetadata;
+
+/**
+ * Default User Metadata Service.
+ *
+ * @author zhenl
+ * @since 1.0.0
```
tgianos: think you mean 1.1.0 here

metacat-common-server/src/main/java/com/netflix/metacat/common/server/properties/Config.java
```diff
@@ -266,4 +266,10 @@
      * @return The partition topic ARN or null if no property set
      */
     String getSnsTopicPartitionArn();
+
+    /**
+     * Use default user metadata service.
+     * @return The flag to turn on mysql based user metadata service.
+     */
+    boolean userMetadataServiceEnabled();
```
tgianos: a comment here about the general idea. there are two ways to do this.

1) Disable the feature completely. So that there is no trace of it in the system at runtime. When it is enabled specify which implementation to use somehow (profile, property, etc)
2) Enable it by default but switch the implementation based on profile, property, etc

This looks like number 2 but the naming of this variable is distinctly pointing towards number 1. I believe we discussed that unfortunately the user metadata service is pretty hardcoded into the application all over the place so 1 may be hard to accomplish but perhaps we should think of doing it.

One way I could see doing it is abstract the entire user metadata behind an adapter. Or the default implementation could really be the adapter. You can give it an optional constructor argument of another implementation and then if that exists at runtime defer to it or if it doesn't return nothing. That way you can still keep the various points in the code where the user metadata service is being called but be able to turn on or off a real implementation. Not sure how others feel or if this is clear.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/properties/User.java
```diff
@@ -28,6 +28,7 @@
  */
 @Data
 public class User {
+    private boolean enabled; // false is default in Java no need to initialize
```
tgianos: Should this be within metadata instead?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/usermetadata/UserMetadataService.java
```diff
@@ -84,7 +92,9 @@
      * @return data metadata for the given uri.
      */
     @Nonnull
```
tgianos: This makes no sense. The entire purpose of returning an Optional is so that you can check if the contents are not null?

Are we just saying the actual Optional itself is nonnull? that's weird.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/usermetadata/UserMetadataService.java
```diff
@@ -93,31 +103,39 @@
...
 
     /**
      * Returns the definition metadata for the given name.
      *
      * @param name name
      * @return definition metadata for the given name
      */
-    Optional<ObjectNode> getDefinitionMetadata(@Nonnull QualifiedName name);
+    default Optional<ObjectNode> getDefinitionMetadata(QualifiedName name) {
+        return Optional.of(JsonNodeFactory.instance.objectNode());
```
tgianos: Why not just return Optional.ofNull instead of returning an actual object? We don't really want anything anyway right?

metacat-main/src/main/java/com/netflix/metacat/main/configs/ServicesConfig.java
```diff
@@ -73,7 +74,9 @@ public UserMetadataService userMetadataService(
         final Config config,
         final MetacatJson metacatJson
     ) {
-        return new MysqlUserMetadataService(dataSourceManager, metacatJson, config);
+        return config.userMetadataServiceEnabled()
```
tgianos: Along with my previous comment this really limits us to ONLY having a mysql or default implementation. Which is not a very good or extensible design. I understand this is the case right now but we should be able to plug in different implementations without modifying this code.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/properties/Config.java
zhljen: The adapter abstraction mentioned above will work. I think the second option, as standard constructor injection, is more proper here. Say we have two beans defined as the implementation of userMetadataService. The default one is the no-op service, which we mark as primary (matchIfMissing). The mysql one can be injected if we set the property explicitly. The property on the profile/property is the name of the implementation, e.g., 'mysql', not 'enabled'.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/usermetadata/DefaultUserMetadataService.java
zhljen: ya, we moved to 1.1.0 now.

metacat-main/src/main/java/com/netflix/metacat/main/configs/ServicesConfig.java
zhljen: Changed to profile based userMetadataService bean injection. This looks to be a very popular  approach to address this case, as we don't have much conditions to consider except picking up one implementation.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/usermetadata/UserMetadataService.java
zhljen: Agreed.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/usermetadata/UserMetadataService.java
```diff
@@ -45,46 +47,52 @@
      *
      * @param uris list of uris.
      */
-    void deleteDataMetadatas(@Nonnull List<String> uris);
+    default void deleteDataMetadatas(@Nonnull List<String> uris) {
```
tgianos: We shouldn't have to put `@Nonnull` on the method parameters if you've put the annotation on the `package-info.java` file as we all agreed.

metacat-user-metadata-mysql/src/main/java/com/netflix/metacat/usermetadata/mysql/MySqlUserMetadataConfig.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.context.annotation.Profile;
+
+/**
+ * MySql UserMetadata Config.
+ *
+ * @author zhenl
+ * @since 1.1.0
+ */
+@Configuration
+@Profile({"dev", "test", "prod"})
```
tgianos: Instead of all these profiles I would just create a profile like "mysql-metadata". We shouldn't have to have all these just to enable this feature. dev, test, prod should probably moreso be reserved for swapping out properties like database connection strings or catalog locations.

metacat-functional-tests/metacat-test-cluster/docker-compose.yml
```diff
@@ -50,6 +50,7 @@ services:
                 -Dnetflix.appinfo.metadata.enableRoute53=false
                 -Dmetacat.plugin.config.location=/etc/metacat/catalog
                 -Dmetacat.elacticsearch.enabled=false
+                -Dspring.profiles.active=test
```
tgianos: If you're doing this to force the mysql-metadata to be on for the tests I would say after you rename the profile be sure to enable that feature here.

zhljen: Agreed.

PR #126, opened 2017-05-25
Add latency for SNS publishing

tgianos:


ajoymajumdar:
Can we instrument the stats of the executor used in the event bus?

tgianos:
We'd probably need to use an aspect to do that. We can talk about that for a different PR. Going to merge this one.

PR #127, opened 2017-05-26
Bug fixes.

ajoymajumdar:
1. SavePartition API should update only table metadata and not the schema when updating metrics.
2. Fixed the usage of threadservicemanager in hive connector.


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/util/HiveConfigConstants.java
```diff
@@ -46,6 +46,10 @@
      * USE_FASTPARTITION_SERVICE.
      */
     public static final String USE_FASTHIVE_SERVICE = "hive.use.embedded.fastservice";
+    /**
+     * USE_FASTPARTITION_SERVICE.
+     */
+    public static final String THREAD_POOL_SIZE = "hive.thread.pool.size";
```
tgianos: if this is some new thing it will have to be added to the master branch separately.

danielbwatson: also the comment needs to be updated

PR #133, opened 2017-06-01
Fix latency in handling events.

ajoymajumdar:
1. Use AmazonSNSAsyncClient instead of AmazonSNSCLient for publishing messages.
2. Log latency for events processed in elastic search and SNS handlers.
3. Fire the table change event before firing the list of partition events.


metacat-main/src/main/java/com/netflix/metacat/main/services/search/MetacatEventHandlers.java
```diff
@@ -104,15 +131,23 @@ public void metacatDeleteDatabasePostEventHandler(final MetacatDeleteDatabasePos
...
-        } catch (Exception e) {
-            log.warn("Failed deleting the partitions for the dropped table/view:{}", dto.getName().toString());
+            final TableDto dto = event.getTable();
+            es.softDelete(ElasticSearchDoc.Type.table.name(), dto.getName().toString(), event.getRequestContext());
+            try {
+                final List<String> partitionIdsToBeDeleted =
+                    es.getIdsByQualifiedName(ElasticSearchDoc.Type.partition.name(), dto.getName());
+                es.delete(ElasticSearchDoc.Type.partition.name(), partitionIdsToBeDeleted);
+            } catch (Exception e) {
+                log.warn("Failed deleting the partitions for the dropped table/view:{}", dto.getName().toString());
```
tgianos: why not error?

PR #136, opened 2017-06-02
Fix latency in handling events.

ajoymajumdar:
Use separate internal event bus for SNS, Kafka and ElasticSearch event handlers.

### DIFF
--- a/metacat-main/src/main/java/com/netflix/metacat/main/init/MetacatInitializationService.java
+++ b/metacat-main/src/main/java/com/netflix/metacat/main/init/MetacatInitializationService.java
@@ -54,7 +54,7 @@ public class MetacatInitializationService {
         this.injector = injector;
 
         // Register all the services to listen for events
-        notificationServices.forEach(eventBus::register);
+        notificationServices.forEach((object) -> eventBus.register(object, "sns"));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


metacat-main/src/main/java/com/netflix/metacat/main/init/MetacatInitializationService.java
```diff
@@ -54,7 +54,7 @@ public MetacatInitializationService(
         this.injector = injector;
 
         // Register all the services to listen for events
-        notificationServices.forEach(eventBus::register);
+        notificationServices.forEach((object) -> eventBus.register(object, "sns"));
```
zhljen: The "sns" should be defined as constant.

metacat-main/src/main/java/com/netflix/metacat/main/init/MetacatInitializationService.java
```diff
@@ -77,7 +77,7 @@ public void start() throws Exception {
             final MetacatEventBus eventBus = injector.getInstance(MetacatEventBus.class);
             // Only register the elastic search event handlers if the client is registered
             final MetacatEventHandlers handlers = injector.getInstance(MetacatEventHandlers.class);
-            eventBus.register(handlers);
+            eventBus.register(handlers, "es");
```
zhljen: Similar as above.

zhljen: We can fix them in the spring change.

PR #138, opened 2017-06-07
Fixed regression issue where updating an existing partition was broken.

ajoymajumdar:
The issue was that the existing partition was not getting updated. Changes were not ported properly from build 0.1.52.


metacat-functional-tests/src/functionalTest/groovy/com/netflix/metacat/MetacatSmokeSpec.groovy
```diff
@@ -69,7 +69,7 @@ class MetacatSmokeSpec extends Specification {
     public static MetacatJson metacatJson = MetacatJsonLocator.INSTANCE
 
     def setupSpec() {
-        String url = "http://localhost:${System.properties['metacat_http_port']}"
+        String url = "http://localhost:32778"
```
tgianos: I don't think you want this port hardcoded

PR #143, opened 2017-06-09
Add missing enable property

tgianos:


zhljen:
Where is this flag used?

tgianos:
In the sea of internal code @zhljen

tgianos:
In the sea of internal code @zhljen

PR #145, opened 2017-06-12
Fix source type generation

tgianos:


ajoymajumdar:
Please add getTable test cases in functionalTest for all tables (TestCatalogs.getAllTables(TestCatalogs.ALL))

PR #147, opened 2017-06-13
Fixed regression issue where setting tags for a redshift table was broken.

ajoymajumdar:
ConnectorBaseService.exists method was not implemented by the Redshift connector.
Fixed regression issue where getTable for non-existing redshift tables were not failing with a 404 error.

### DIFF
--- a/metacat-connector-jdbc/src/main/java/com/netflix/metacat/connector/jdbc/services/JdbcConnectorTableService.java
+++ b/metacat-connector-jdbc/src/main/java/com/netflix/metacat/connector/jdbc/services/JdbcConnectorTableService.java
@@ -110,10 +110,7 @@ public class JdbcConnectorTableService implements ConnectorTableService {
     @Override
     public TableInfo get(@Nonnull final ConnectorContext context, @Nonnull final QualifiedName name) {
         log.debug("Beginning to get table metadata for qualified name {} for request {}", name, context);
-        // If table does not exist, throw TableNotFoundException.
-        if (!exists(context, name)) {
-            throw new TableNotFoundException(name);
-        }
+
         try (Connection connection = this.dataSource.getConnection()) {
             final String database = name.getDatabaseName();
             connection.setSchema(database);
@@ -138,6 +135,11 @@ public class JdbcConnectorTableService implements ConnectorTableService {
 
                     fields.add(fieldInfo.build());
                 }
+            }
+            final List<FieldInfo> fieldInfos = fields.build();
+            // If table does not exist, throw TableNotFoundException.
+            if (fieldInfos.isEmpty() && !exists(context, name)) {
+                throw new TableNotFoundException(name);
             }
             log.debug("Finished getting table metadata for qualified name {} for request {}", name, context);
             return TableInfo.builder().name(name).fields(fields.build()).build();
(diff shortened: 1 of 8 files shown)


metacat-connector-jdbc/src/main/java/com/netflix/metacat/connector/jdbc/services/JdbcConnectorTableService.java
```diff
@@ -94,15 +95,9 @@ public void delete(@Nonnull final ConnectorContext context, @Nonnull final Quali
...
-        try (final Connection connection = this.dataSource.getConnection()) {
-            final boolean upperCase = connection.getMetaData().storesUpperCaseIdentifiers();
-            if (upperCase) {
-                connection.setSchema(databaseName.toUpperCase(Locale.ENGLISH));
-            } else {
-                connection.setSchema(databaseName);
-            }
-            final String finalTableName = upperCase ? tableName.toUpperCase(Locale.ENGLISH) : tableName;
-            JdbcConnectorUtils.executeUpdate(connection, this.getDropTableSql(name, finalTableName));
+        try (Connection connection = this.dataSource.getConnection()) {
```
tgianos: Why did you remove the `final` from these try blocks?

metacat-connector-jdbc/src/main/java/com/netflix/metacat/connector/jdbc/services/JdbcConnectorTableService.java
```diff
@@ -262,6 +253,23 @@ public void rename(
         }
     }
 
+    @Override
+    public boolean exists(@Nonnull final ConnectorContext context, @Nonnull final QualifiedName name) {
+        boolean result = false;
```
tgianos: this is the default value of a boolean anyway you don't need to initialized it.

metacat-connector-jdbc/src/main/java/com/netflix/metacat/connector/jdbc/services/JdbcConnectorTableService.java
```diff
@@ -262,6 +253,23 @@ public void rename(
...
+        try (Connection connection = this.dataSource.getConnection()) {
+            final String databaseName = name.getDatabaseName();
+            connection.setSchema(databaseName);
+            final DatabaseMetaData metaData = connection.getMetaData();
+            final ResultSet rs = metaData.getTables(databaseName, databaseName, name.getTableName(), TABLE_TYPE);
+            if (rs.next()) {
```
tgianos: here it might be better to do `rs.last()` and then `return rs.row() == 1`. That way we verify that the table exists and it's the only one.

metacat-connector-jdbc/src/main/java/com/netflix/metacat/connector/jdbc/services/JdbcConnectorTableService.java
ajoymajumdar: Not needed in try-with-resources statement.

metacat-connector-jdbc/src/main/java/com/netflix/metacat/connector/jdbc/services/JdbcConnectorTableService.java
ajoymajumdar: At a method level, you need to initialize.

metacat-connector-jdbc/src/main/java/com/netflix/metacat/connector/jdbc/services/JdbcConnectorTableService.java
ajoymajumdar: Using next() should be acceptable with the assumption that databases will not have two tables with the same name.

PR #160, opened 2017-06-28
Code cleanup NonNull related changes

zhljen:


metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatUpdateTablePreEvent.java
```diff
@@ -51,8 +52,8 @@ public MetacatUpdateTablePreEvent(
         @Nonnull @NonNull final QualifiedName name,
         @Nonnull @NonNull final MetacatRequestContext requestContext,
         @Nonnull @NonNull final Object source,
-        @Nonnull @NonNull final TableDto oldTable,
-        @Nonnull @NonNull final TableDto currentTable
+        @Nullable final TableDto oldTable,
```
tgianos: Now that these are nullable you basically need to go to every place these events are used and check for null too...

zhljen:
After discussion, I reverted this change and marked as TODO.

zhljen:
Revert back the !=null removals.

metacat-common/src/main/java/com/netflix/metacat/common/dto/BaseDto.java
```diff
@@ -35,7 +35,7 @@
  */
 //TODO: All DTO's should be READ-ONLY
 public abstract class BaseDto implements Serializable {
-    private static MetacatJsonLocator metacatJsonLocator = new MetacatJsonLocator();
+    protected static final MetacatJsonLocator METACAT_JSON_LOCATOR = new MetacatJsonLocator();
```
tgianos: any reason this is static rather than a member?

zhljen: Don't want to create instances in the deserializeObjectNode/ serializeObjectNode BaseDTO.

PR #161, opened 2017-06-29
Add health indicators for internal services and catalogs

tgianos:
This PR adds implementations of the `HealthIndicator` interface for the `MetacatInititializationService` and `CatalogManager`. This way the system won't report `UP` via the actuator `/health` endpoint until these services are considered up.

This will also control whether the node is up or down in discovery if there is Eureka integration.


ajoymajumdar:
I would suggest creating a class that will implement HealthIndicator and checking MetacatInitializationService.

PR #167, opened 2017-07-18
Not use the default ObjectMapper from Spring.

ajoymajumdar:
Also fixed issue where the Serde inputFormat was getting set to empty string instead of NULL.

### DIFF
--- a/metacat-main/src/main/java/com/netflix/metacat/main/configs/CommonServerConfig.java
+++ b/metacat-main/src/main/java/com/netflix/metacat/main/configs/CommonServerConfig.java
@@ -17,6 +17,9 @@
  */
 package com.netflix.metacat.main.configs;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.netflix.metacat.common.json.MetacatJson;
 import com.netflix.metacat.common.json.MetacatJsonLocator;
 import com.netflix.metacat.common.server.converter.ConverterUtil;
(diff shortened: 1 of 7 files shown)


metacat-main/src/main/java/com/netflix/metacat/main/configs/CommonServerConfig.java
```diff
@@ -48,6 +53,20 @@
 public class CommonServerConfig {
 
     /**
+     * An object mapper bean to use if none already exists.
+     *
+     * @return JSON object mapper
+     */
+    @Primary
```
tgianos: Rather than doing all this with a primary bean and whatnot is there any reason you didn't use the properties to configure the spring one?

```
spring.jackson.date-format= # Date format string or a fully-qualified date format class name. For instance `yyyy-MM-dd HH:mm:ss`.
spring.jackson.default-property-inclusion= # Controls the inclusion of properties during serialization.
spring.jackson.deserialization.*= # Jackson on/off features that affect the way Java objects are deserialized.
spring.jackson.generator.*= # Jackson on/off features for generators.
spring.jackson.joda-date-time-format= # Joda date time format string. If not configured, "date-format" will be used as a fallback if it is configured with a format string.
spring.jackson.locale= # Locale used for formatting.
spring.jackson.mapper.*= # Jackson general purpose on/off features.
spring.jackson.parser.*= # Jackson on/off features for parsers.
spring.jackson.property-naming-strategy= # One of the constants on Jackson's PropertyNamingStrategy. Can also be a fully-qualified class name of a PropertyNamingStrategy subclass.
spring.jackson.serialization.*= # Jackson on/off features that affect the way Java objects are serialized.
spring.jackson.time-zone= # Time zone used when formatting dates. For instance `America/Los_Angeles`
```

PR #186, opened 2017-08-29
Populate restApiuri and scheme in metacatContextRequst

zhljen:


metacat-common/src/main/java/com/netflix/metacat/common/MetacatRequestContext.java
```diff
@@ -65,18 +73,21 @@
...
+     * @param apiPathPrefix   the path prefix of rest api
      */
     public MetacatRequestContext(
         @Nullable final String userName,
         @Nullable final String clientAppName,
         @Nullable final String clientId,
         @Nullable final String jobId,
-        @Nullable final String dataTypeContext
+        @Nullable final String dataTypeContext,
+        @Nullable final String apiPathPrefix
```
tgianos: This should just be API version and it should probably be an integer. If you intend to capture something like `thrift` vs. `http` then you're talking about some sort of invocation scheme. I'd recommend breaking it up into two fields for clarity to save parsing later on as more API versions or layers cold be added.

metacat-main/src/main/java/com/netflix/metacat/main/api/ApiFilter.java
```diff
@@ -65,14 +65,21 @@ public void doFilter(
         final String clientHost = httpServletRequest.getHeader("X-Forwarded-For");
         final String jobId = httpServletRequest.getHeader(MetacatRequestContext.HEADER_KEY_JOB_ID);
         final String dataTypeContext = httpServletRequest.getHeader(MetacatRequestContext.HEADER_KEY_DATA_TYPE_CONTEXT);
+        String apiVersion = null;
+        for (Object version: MetacatRequestContext.REST_API_PATH_PREFIX) {
```
tgianos: Why is this an object and not a String?

metacat-main/src/main/java/com/netflix/metacat/main/api/ApiFilter.java
```diff
@@ -65,14 +65,21 @@ public void doFilter(
...
+            if (httpServletRequest.getRequestURI().startsWith((String) version)) {
+                apiVersion = (String) version;
```
tgianos: I think this is pretty inefficient to do this on every request when the request is already being filtered and checked by the web framework. I think a possible cleaner design would be to instantiate multiple instances of this filter and bind them to the different paths and set the `apiVersion` in the constructor and use it here as a constant. That way once the filter is invoked you're already in the right path. Simply put you could have two filters one instantiated for `0` listening on `/mds/v0/*` and one instantiated for `1` listening on `/mds/v1/*` each would only be invoked when those paths are requested. No need to do extra checking on the path. That would be my suggestion or something similar that doesn't involve so much coding of constants and instead relies on configuration and re-use.

tgianos: As a total aside are we tied to this whole `/mds` thing? when we deprecate `v0` can we move it to something like `/api`? Might just be my OCD talking though haha

tgianos:
@ajoymajumdar as a total aside are we tied to this whole `/mds` thing? When we deprecate `v0` can we move it to something like `/api`? Might just be my OCD talking though.

tgianos:
@zhljen you need to rebase this PR

zhljen:
I think it through and decide to capture the whole request uri and scheme ( http/thrft) in the context, i.e. just two extra get calls in the filter. Later on, we will put more complex logic in the retrieval part to get the version, path, scheme. I tried to publish them in keystone and the current keystone publisher is an event listener, which is an event listener and won't have the big impact to the whole performance comparing to the filter.

metacat-common/src/main/java/com/netflix/metacat/common/MetacatRequestContext.java
```diff
@@ -48,14 +51,22 @@
...
     private final String id = UUID.randomUUID().toString();
     // TODO: Move to Java 8 and use java.time.Instant
     private final long timestamp = new Date().getTime();
-    private final String userName;
-    private final String clientAppName;
-    private final String clientId;
-    private final String jobId;
-    private final String dataTypeContext;
+
+    private String userName;
```
tgianos: These variables should all be final as the request context should be immutable.

zhljen:
Changed metacatContextRequest to immutable and updated the PR.

metacat-common/src/main/java/com/netflix/metacat/common/MetacatRequestContext.java
```diff
@@ -65,18 +87,141 @@
...
+     * @return the builder class for MetacatRequestContext
+     */
+    public static MetacatRequestContext.MetacatRequestContextBuilder builder() {
+        return new MetacatRequestContext.MetacatRequestContextBuilder();
+    }
+
+    /**
+     * MetacatRequestContext builder class.
+     */
+    public static class MetacatRequestContextBuilder {
```
tgianos: I'm pretty sure there's a bug in here. Since the `apiUri` and the `scheme` can't be null in the actual instance this builder has no required parameters or default settings for those variables. If the setters aren't called the fields in this builder remain `null` and when the `build` method is called nulls will be passed to the parameterized constructor above. In the default constructor below you should probably assign your api and scheme to whatever default values you wanted.

metacat-common/src/main/java/com/netflix/metacat/common/MetacatRequestContext.java
```diff
@@ -65,18 +87,141 @@
...
+        private String buserName;
+        private String bclientAppName;
```
tgianos: the camel casing here doesn't look right. Shouldn't it be `bUserName`?

PR #187, opened 2017-09-05
upgrading elasticsearch to 5.x.

zhljen:


build.gradle
```diff
@@ -138,6 +138,10 @@ configure(javaProjects) {
             dependency("org.glassfish.jersey.containers:jersey-container-servlet:2.19")
             dependency("org.glassfish.jersey.media:jersey-media-json-jackson:2.19")
             dependency("org.spockframework:spock-guice:1.0-groovy-2.4")
+            /**es 5.4.1 dependencies*/
+            dependency("org.elasticsearch.client:transport:5.4.1")
+            dependency("org.apache.logging.log4j:log4j-api:2.7")
```
tgianos: Why do we need these explicitly here?

metacat-main/build.gradle
```diff
@@ -29,9 +29,12 @@ dependencies {
     compile(project(":metacat-common"))
     compile(project(":metacat-common-server"))
     compile(project(":metacat-thrift"))
-    compile("org.elasticsearch:elasticsearch")
+    compile("org.elasticsearch.client:transport")
+
     compile("com.amazonaws:aws-java-sdk-sns")
     compile("joda-time:joda-time")
+    compile("org.apache.logging.log4j:log4j-api")
```
tgianos: Is there a reason we're using log4j like this here instead of using slf4j/logback like everywhere else?

metacat-main/src/main/java/com/netflix/metacat/main/configs/ElasticSearchConfig.java
```diff
@@ -27,34 +27,41 @@
...
+ * @author zhenl
  * @since 1.1.0
  */
-//TODO: Look into spring data elasticsearch to replace this
+@Slf4j
 @Configuration
 @ConditionalOnProperty(value = "metacat.elasticsearch.enabled", havingValue = "true")
 public class ElasticSearchConfig {
 
+    private static final String TCP_CONNECTION_TIMEOUT = "60s";
```
tgianos: With this being used only once maybe you should just have it as a local variable? Additionally this looks like something that would want to be a property with a default value.

build.gradle
zhljen: To pin the version of elasticSearch client instead of relying on the default one.  I removed the dependency of org.apache.logging.log4j:log4j-api:2.7.

metacat-main/src/main/java/com/netflix/metacat/main/configs/ElasticSearchConfig.java
zhljen: Ok. Don't think we need to make it configurable. Roll back the code to previous.

metacat-main/build.gradle
ajoymajumdar: Then why do we have to explicitly depend on it. It does not get it transitively

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchDoc.java
```diff
@@ -51,11 +53,13 @@
 
     private String id;
     private Object dto;
+    private final String timestamp = String.valueOf(Instant.now().toEpochMilli());
```
ajoymajumdar: Can we set the timestamp in the constructor and have this as null by default

metacat-main/build.gradle
zhljen: It may change based on the spring boot version.
want to set it to a specific version, which the cluster is.

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchDoc.java
zhljen: Why do we want to put null as default? This field is always required.

ajoymajumdar: When we read the document from elasticsearch, how will you populate this Object?

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchEventHandlers.java
```diff
@@ -250,9 +249,9 @@ private void updateEntitiesWithSameUri(final String metadataType, final TableDto
             final List<String> ids = es.getTableIdsByUri(metadataType, dto.getDataUri());
             ids.remove(dto.getName().toString());
             if (!ids.isEmpty()) {
-                final ObjectNode node = metacatJsonLocator.emptyObjectNode();
-                node.set("dataMetadata", dto.getDataMetadata());
-                es.updates(ElasticSearchDoc.Type.table.name(), ids, metacatRequestContext, node);
+               final ElasticSearchDoc doc = new ElasticSearchDoc(dto.getName().toString(), dto,
```
ajoymajumdar: Is this correct?

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchDoc.java
zhljen: I am not aware of this use-case.Needs a bit explain how to use it after reading from the es. If we need to update a certain field, we can upload the fields explicitly.

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchEventHandlers.java
zhljen: I am not sure what the previous code is doing to be honest. In the new world, the API does not take in the node, we can talk about it and update.

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchUtilImpl.java
```diff
@@ -90,13 +90,14 @@
...
-     * @param config      config
-     * @param metacatJson json utility
-     * @param elasticSearchMetric    elastic search metric
+     * @param client              elastic search client
+     * @param config              config
+     * @param metacatJson         json utility
+     * @param elasticSearchMetric elastic search metric
      */
     public ElasticSearchUtilImpl(
         @Nullable final Client client,
```
ajoymajumdar: Does the new ES client have retry logic? If so, can we use that instead of us handling the retry logic.

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchUtilImpl.java
```diff
@@ -646,35 +641,44 @@ private void softDeleteDoc(
...
+            log.error("Failed saving metadata of index {} type {} with id {}: {}. {}",
+                index, type, id, doc.getDto(), e);
             this.elasticSearchMetric.getElasticSearchSaveFailureCounter().increment();
             log("ElasticSearchUtil.saveToIndex", type, id, null, e.getMessage(), e, true, index);
         }
     }
 
     private static List<String> getIds(final SearchResponse response) {
-        return FluentIterable.from(response.getHits()).transform(SearchHit::getId).toList();
+        final List<String> ret = Lists.newArrayList();
```
ajoymajumdar: Why cant we use the Iterable?

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchUtilImpl.java
```diff
@@ -646,35 +641,44 @@ private void softDeleteDoc(
...
+        for (SearchHit hit : response.getHits().getHits()) {
+            ret.add(hit.getId());
+        }
+        return ret;
     }
 
     private <T> List<T> parseResponse(final SearchResponse response, final Class<T> valueType) {
-        return FluentIterable.from(response.getHits()).transform(hit -> {
+        final List<T> ret = Lists.newArrayList();
```
ajoymajumdar: Why cant we use the Iterable?

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchEventHandlers.java
zhljen: Updated the logic to previous implementation.

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchDoc.java
zhljen: Agreed. Moved timestamp init to constructor.

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchUtilImpl.java
zhljen: No it doesn't have the retry logic.

metacat-main/build.gradle
```diff
@@ -74,14 +75,16 @@ dependencies {
     runtime("io.springfox:springfox-swagger-ui")
     runtime("org.webjars:hal-browser")
     runtime("org.springframework:spring-aspects")
+    runtime("org.apache.logging.log4j:log4j-api")
```
ajoymajumdar: This can be removed since log4j-core gets log4j-api

zhljen:
Bring back the previous elasticSearch using embedded es server.

metacat-main/build.gradle
```diff
@@ -74,14 +75,16 @@ dependencies {
...
+    runtime("org.apache.logging.log4j:log4j-core")
 
     /*******************************
      * Test Dependencies
      *******************************/
 
     testCompile(project(":metacat-testdata-provider"))
-
     testCompile("io.airlift:testing-mysql-server")
+    testCompile("org.apache.logging.log4j:log4j-core")
```
ajoymajumdar: Can we change this to testRuntime?

zhljen: Yes, tried that. The build process will fail using testRuntime
:metacat-main:compileTestGroovy
> Unable to load class org.elasticsearch.node.Node due to missing dependency org/apache/logging/log4j/Logger

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchEventHandlers.java
```diff
@@ -236,23 +237,25 @@ public void metacatUpdateTablePostEventHandler(final MetacatUpdateTablePostEvent
...
-                                           final MetacatRequestContext metacatRequestContext) {
+                                           final String userName) {
         if (dto.isDataExternal()) {
             final List<String> ids = es.getTableIdsByUri(metadataType, dto.getDataUri());
             ids.remove(dto.getName().toString());
             if (!ids.isEmpty()) {
                 final ObjectNode node = metacatJsonLocator.emptyObjectNode();
-                node.set("dataMetadata", dto.getDataMetadata());
-                es.updates(ElasticSearchDoc.Type.table.name(), ids, metacatRequestContext, node);
+                node.set(ElasticSearchDoc.Field.DATA_METADATA, dto.getDataMetadata());
```
ajoymajumdar: why are we not setting the timestamp?

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchUtilImpl.java
```diff
@@ -90,13 +90,14 @@
...
-     * @param client      elastic search client
-     * @param config      config
-     * @param metacatJson json utility
-     * @param elasticSearchMetric    elastic search metric
+     * @param client              elastic search client
+     * @param config              config
+     * @param metacatJson         json utility
+     * @param elasticSearchMetric elastic search metric
      */
     public ElasticSearchUtilImpl(
```
ajoymajumdar: We need to set the timestamp on softDelete, update and create.

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchUtilImpl.java
```diff
@@ -371,7 +355,7 @@ private ObjectNode toJsonObject(final ElasticSearchDoc elasticSearchDoc) {
         final QueryBuilder queryBuilder = QueryBuilders.boolQuery()
             .must(QueryBuilders.termsQuery("name.qualifiedName.tree", names))
             .must(QueryBuilders.termQuery("deleted_", false))
-            .must(QueryBuilders.rangeQuery("_timestamp").lte(marker.toDate()))
+            .must(QueryBuilders.rangeQuery(ElasticSearchDoc.Field.TIMESTAMP).lte(marker.toDate()))
```
ajoymajumdar: Does this work? We are storing the timestamp in epoch millis.

zhljen: I will update to milli for comparison.

ajoymajumdar: Please add a test case

metacat-main/src/test/resources/search/mapping/metacat.json
```diff
@@ -1,6 +1,7 @@
 {
     "settings": {
-        "index":{"number_of_shards":10},
+        "index":{"number_of_shards":5,
+                 "max_result_window": 2147483647},
```
ajoymajumdar: Why are we setting this property high?

zhljen: We discussed yesterday, it's needed as setting the size in the query.
 final SearchRequestBuilder request = client.prepareSearch(esIndex)
            .setTypes(type)
            .setSearchType(SearchType.QUERY_THEN_FETCH)
            .setQuery(queryBuilder)
            .setSize(Integer.MAX_VALUE);

metacat-main/src/test/resources/search/mapping/metacat.json
```diff
@@ -15,71 +16,47 @@
...
                         "inputFormat": {
-                            "type": "string",
-                            "fields": {
-                                "raw": {
-                                    "type": "string",
-                                    "index": "not_analyzed"
-                                }
-                            }
+                            "type": "keyword",
+                            "index": true
```
ajoymajumdar: We should set this to:
{
  "inputFormat": {
    "type" "text",
   "index" = true,
    "fields": {
      "keyword": {
        "type": "keyword",
        "ignore_above": 256
      }
    }
  }
}

ajoymajumdar: Same applies to outputformat, owner, uri and serializationLib

metacat-main/src/test/resources/search/mapping/metacat.json
```diff
@@ -154,59 +131,59 @@
...
                         }
                     }
                 },
+                "timestamp": {
+                  "type": "date",
+                  "format": "epoch_millis",
+                  "index": true
+                },
                 "user_": {
-                    "type": "string",
```
ajoymajumdar: Need keyword and text like inputFormat

metacat-main/src/test/resources/search/mapping/metacat.json
```diff
@@ -154,59 +131,59 @@
...
                         "qualifiedName": {
-                            "type": "string",
-                            "index": "not_analyzed",
+                            "type": "keyword",
+                            "index": true,
                             "fields": {
                                 "tree": {
-                                    "type": "string",
+                                    "type": "text",
                                     "analyzer": "paths"
```
ajoymajumdar: Do we have to say index=true for the property 'tree'.

zhljen:
Updated the mapping based on comments.

PR #188, opened 2017-09-06
Fixed issue where all events were incorrectly handled asynchronously. 

ajoymajumdar:
Pre-events needed to be handled synchronously. Also made other minor fixes.

- Added &rewriteBatchedStatements=true to mysql jdbc url to enable batch updates
- Removed the usage of List.remove in ElasticSearchEventHandlers.

### DIFF
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationEventMulticaster.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationEventMulticaster.java
@@ -1,0 +1,91 @@
+/*
+ *
+ *  Copyright 2016 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.metacat.common.server.events;
+
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.context.ApplicationEvent;
+import org.springframework.context.ApplicationListener;
+import org.springframework.context.event.ApplicationEventMulticaster;
+import org.springframework.context.event.SimpleApplicationEventMulticaster;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Event bus implementation using Springs Event Multicaster. By default, Spring supports sunchronous event publishing.
+ * This implementation supports both synchronous and asynchronous event publishing. At present, client explicitly
+ * call postAsync or postSync to appropriately publish events.
+ * This implementation should be used as the application event multicaster in the Springs context.
+ * Synchronous publishing of events is handled by this class and the asynchronous publishing of events is handled by
+ * the asyncEventMulticaster.
+ *
+ * @author amajumdar
+ * @since 1.1.x
+ */
+@Slf4j
+public class MetacatApplicationEventMulticaster extends SimpleApplicationEventMulticaster {
+
+    private final ApplicationEventMulticaster asyncEventMulticaster;
+    /**
+     * Constructor.
+     *
+     * @param asyncEventMulticaster The asynchronous event multicaster to use
+     */
+    public MetacatApplicationEventMulticaster(
+        @Nonnull @NonNull final ApplicationEventMulticaster asyncEventMulticaster
+    ) {
+        super();
+        this.asyncEventMulticaster = asyncEventMulticaster;
+    }
+
+    /**
+     * Post event asynchronously.
+     *
+     * @param event event
+     */
+    public void postAsync(final ApplicationEvent event) {
+        this.asyncEventMulticaster.multicastEvent(event);
+    }
+
+    /**
+     * Post event synchronously.
+     *
+     * @param event event
+     */
+    public void postSync(final ApplicationEvent event) {
+        super.multicastEvent(event);
+    }
+
+    @Override
+    public void addApplicationListener(final ApplicationListener listener) {
+        super.addApplicationListener(listener);
+        asyncEventMulticaster.addApplicationListener(listener);
+    }
+
+    @Override
+    public void removeApplicationListener(final ApplicationListener listener) {
+        super.removeApplicationListener(listener);
+        asyncEventMulticaster.removeApplicationListener(listener);
+    }
+
+    @Override
+    public void removeAllListeners() {
+        super.removeAllListeners();
+        asyncEventMulticaster.removeAllListeners();
+    }
+}
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
@@ -23,8 +23,6 @@ import com.netflix.spectator.api.Registry;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.context.ApplicationEvent;
-import org.springframework.context.ApplicationEventPublisher;
-import org.springframework.context.event.ApplicationEventMulticaster;
 
 import javax.annotation.Nonnull;
 
@@ -38,24 +36,20 @@ import javax.annotation.Nonnull;
 @Slf4j
 public class MetacatEventBus {
 
-    private final ApplicationEventPublisher eventPublisher;
-    private final ApplicationEventMulticaster eventMulticaster;
+    private final MetacatApplicationEventMulticaster applicationEventMulticaster;
     private final Counter eventAsyncCounter;
     private final Counter eventSyncCounter;
     /**
      * Constructor.
      *
-     * @param eventPublisher   The synchronous event publisher to use
-     * @param eventMulticaster The asynchronous event multicaster to use
+     * @param applicationEventMulticaster The event multicaster to use
      * @param registry         The registry to spectator
      */
     public MetacatEventBus(
-            @Nonnull @NonNull final ApplicationEventPublisher eventPublisher,
-            @Nonnull @NonNull final ApplicationEventMulticaster eventMulticaster,
-            @Nonnull @NonNull final Registry registry
+        @Nonnull @NonNull final MetacatApplicationEventMulticaster applicationEventMulticaster,
+        @Nonnull @NonNull final Registry registry
     ) {
-        this.eventPublisher = eventPublisher;
-        this.eventMulticaster = eventMulticaster;
+        this.applicationEventMulticaster = applicationEventMulticaster;
         this.eventAsyncCounter = registry.counter(Metrics.CounterEventAsync.getMetricName());
         this.eventSyncCounter = registry.counter(Metrics.CounterEventSync.getMetricName());
     }
@@ -68,7 +62,7 @@ public class MetacatEventBus {
     public void postAsync(final ApplicationEvent event) {
         log.debug("Received request to post an event {} asynchronously", event);
         this.eventAsyncCounter.increment();
-        this.eventMulticaster.multicastEvent(event);
+        this.applicationEventMulticaster.postAsync(event);
     }
 
     /**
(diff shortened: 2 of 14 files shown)


metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
```diff
@@ -36,26 +37,23 @@
  * @since 0.x
  */
 @Slf4j
-public class MetacatEventBus {
+public class MetacatEventBus extends SimpleApplicationEventMulticaster {
```
tgianos: A concern I have with this approach is the tight coupling this introduces to this Spring class. It's one thing to change how the event bus is created (constructor) it's another to extend this class because now we're exposing a bunch of API's that are only in Spring and could change and we're not in control of them anymore.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
```diff
@@ -36,26 +37,23 @@
...
      *
-     * @param eventPublisher   The synchronous event publisher to use
-     * @param eventMulticaster The asynchronous event multicaster to use
+     * @param asyncEventMulticaster The asynchronous event multicaster to use
      * @param registry         The registry to spectator
      */
     public MetacatEventBus(
-            @Nonnull @NonNull final ApplicationEventPublisher eventPublisher,
-            @Nonnull @NonNull final ApplicationEventMulticaster eventMulticaster,
+            @Nonnull @NonNull final ApplicationEventMulticaster asyncEventMulticaster,
```
tgianos: If you get rid of the synchronous one how are you sending the synchronous events?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
```diff
@@ -79,22 +77,29 @@ public void postAsync(final ApplicationEvent event) {
...
     }
 
-    /**
-     * Registers an object.
-     *
-     * @param object object
-     */
-    public void register(final Object object) {
+    @Override
+    public void addApplicationListener(final ApplicationListener listener) {
```
tgianos: So instead of the extra code with annotations like we saw in that example today instead what you're doing here is registering all listeners in both classes and then based on the API called publishing to one or the other. Are you sure the `super.multicastEvent(event)` call is going to use the default synchronous task executor? I just worry this is a house of cards relying on too much assumptions in configuration that we could screw up at any time or other starters could rather than making it explicit

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
```diff
@@ -79,22 +77,29 @@ public void postAsync(final ApplicationEvent event) {
...
     }
 
-    /**
-     * De-registers an object.
-     *
-     * @param object object
-     */
-    public void unregister(final Object object) {
+    @Override
+    public void removeApplicationListener(final ApplicationListener listener) {
```
tgianos: You should at least put comments as to why you're doing this for future reference

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
```diff
@@ -79,22 +77,29 @@ public void postAsync(final ApplicationEvent event) {
...
+    }
+
+    @Override
+    public void removeAllListeners() {
+        super.removeAllListeners();
+        asyncEventMulticaster.removeAllListeners();
+    }
+
+    @Override
+    public void multicastEvent(final ApplicationEvent event) {
```
tgianos: Is there a point to having this in here? If you don't put this method in here doesn't it just call the super anyway

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorPartitionService.java
```diff
@@ -268,7 +268,7 @@ public PartitionsSaveResponse savePartitions(
             // Existing partition map
             Map<String, Partition> existingPartitionMap = Collections.emptyMap();
 
-            if (partitionsSaveRequest.getCheckIfExists()) {
+            if (partitionsSaveRequest.getCheckIfExists() || partitionsSaveRequest.getAlterIfExists()) {
```
tgianos: I'm going to channel @mprimi here... These changes should at least be in a different commit as you're changing behavior (not just code formatting) and so if we ever go to search the git history this will be totally lost as we think this commit will only be related to the events. You should probably do a `git reset --soft HEAD~1` on your local branch and then commit each of the changes as separate commits with different messages. Then instead of squash merging we can rebase and fast-forward merge and keep the commits separate.

metacat-functional-tests/metacat-test-cluster/etc-metacat/catalog/embedded-hive-metastore.properties
```diff
@@ -50,7 +50,7 @@ hive.metastore.fs.handler.class=com.netflix.metacat.connector.hive.metastore.Hiv
 
 javax.jdo.option.name=embedded-hive-metastore
 javax.jdo.option.username=metacat_user
-javax.jdo.option.url=jdbc:mysql://hive-metastore-db:3306/hive?createDatabaseIfNotExist=true&useUnicode=true&characterEncoding=latin1&autoReconnect=true
+javax.jdo.option.url=jdbc:mysql://hive-metastore-db:3306/hive?createDatabaseIfNotExist=true&useUnicode=true&characterEncoding=latin1&autoReconnect=true&rewriteBatchedStatements=true
```
tgianos: separate commit

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchEventHandlers.java
```diff
@@ -247,8 +247,8 @@ public void metacatUpdateTablePostEventHandler(final MetacatUpdateTablePostEvent
     private void updateEntitiesWithSameUri(final String metadataType, final TableDto dto,
                                            final MetacatRequestContext metacatRequestContext) {
         if (dto.isDataExternal()) {
-            final List<String> ids = es.getTableIdsByUri(metadataType, dto.getDataUri());
-            ids.remove(dto.getName().toString());
+            final List<String> ids = es.getTableIdsByUri(metadataType, dto.getDataUri())
+                .stream().filter(s -> !s.equals(dto.getName().toString())).collect(Collectors.toList());
```
tgianos: is `getName()` not already a string? Seems like the call to `toString()` is redundant but I'm not sure of the return type.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
ajoymajumdar: The reason is how event handling works in Spring. We need MetacatEventBus to either implement the interface or extend the basic implementation. I prefer the later because it has other implementations that I am not clear about. Hence I left it as it is and made it the default event multicaster.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
ajoymajumdar: MetacatEventBus is the default event publisher that processes events synchronously. I do not want to create another event publisher within it.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
ajoymajumdar: Will do.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
tgianos: I understand that I'm just still in favor of somehow doing this outside of manipulating the EventBus in this manner and somehow having whatever is sent into the EventBus be more correct. That way we don't really have to touch the event bus core logic code all the time...

metacat-functional-tests/metacat-test-cluster/etc-metacat/catalog/mysql-56-db.properties
```diff
@@ -19,7 +19,7 @@ connector.name=mysql
 metacat.schema.list-views-with-tables=true
 
 javax.jdo.option.name=mysql-56-db
-javax.jdo.option.url=jdbc:mysql://hive-metastore-db:3306/?useUnicode=true&characterEncoding=latin1&autoReconnect=true
+javax.jdo.option.url=jdbc:mysql://hive-metastore-db:3306/?useUnicode=true&characterEncoding=latin1&autoReconnect=true&rewriteBatchedStatements=true
```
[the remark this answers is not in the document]
ajoymajumdar: Yes, we need to check.
In Datanucleus, the default batching is 50 statements. I would like to increase it but we will limit it based on the max_allowed_packet(4MB on the server). Also usermetadata, we batch it to 2500. We need to check that too. I added this in the tests to check if any functionality is broken.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationEventMulticaster.java
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * Post event synchronously.
+     *
+     * @param event event
+     */
+    public void postSync(final ApplicationEvent event) {
+        super.multicastEvent(event);
+    }
+
+    @Override
```
tgianos: You should add `{@inheritdoc}` to these.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationEventMulticaster.java
```diff
@@ -0,0 +1,99 @@
...
+        super.removeAllListeners();
+        asyncEventMulticaster.removeAllListeners();
+    }
+
+    /**
+     * This is a no op. Use methods, postSync or postAsync.
+     * @param event application event
+     */
+    @Override
+    public void multicastEvent(final ApplicationEvent event) {
```
tgianos: What happens if somehow (misconfiguration) this implementation of `SimpleApplicationEventMulticaster` gets used by the framework itself somewhere and then Spring tries to send an event. I worry this will break framework behavior.

ajoymajumdar: Actually you are right, I should remove this.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
```diff
@@ -38,24 +35,20 @@
...
-     * @param eventPublisher   The synchronous event publisher to use
-     * @param eventMulticaster The asynchronous event multicaster to use
+     * @param applicationEventMulticaster The event multicaster to use
      * @param registry         The registry to spectator
      */
     public MetacatEventBus(
-            @Nonnull @NonNull final ApplicationEventPublisher eventPublisher,
-            @Nonnull @NonNull final ApplicationEventMulticaster eventMulticaster,
-            @Nonnull @NonNull final Registry registry
+        @Nonnull final MetacatApplicationEventMulticaster applicationEventMulticaster,
```
tgianos: Didn't we want to put these `@Nonnull` at the package level and leave in the Lombok ones cause they're runtime catches?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
```diff
@@ -79,22 +72,6 @@ public void postAsync(final ApplicationEvent event) {
     public void postSync(final ApplicationEvent event) {
         log.debug("Received request to post an event {} synchronously", event);
         this.eventSyncCounter.increment();
```
tgianos: Improvement for later: We should tag these metrics with the type of the event. This will allow us to track what's going where and the scale of them. @zhljen has done a lot of this

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationEventMulticaster.java
```diff
@@ -0,0 +1,90 @@
...
+ * call postAsync or postSync to appropriately publish events.
+ * This implementation should be used as the application event multicaster in the Springs context.
+ * Synchronous publishing of events is handled by this class and the asynchronous publishing of events is handled by
+ * the asyncEventMulticaster.
+ *
+ * @author amajumdar
+ * @since 1.1.x
+ */
+@Slf4j
+public class MetacatApplicationEventMulticaster extends SimpleApplicationEventMulticaster {
```
tgianos: There don't seem to be any tests for this class? I think we should have some no?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
ajoymajumdar: Yeah I will revert this change. If I have to do it at the package level, a lot of files need to change. Will do it in another PR.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationEventMulticaster.java
ajoymajumdar: Yes. We have functional tests that covers these.

PR #189, opened 2017-09-22
Refactored the code around spring transaction semantics. 

ajoymajumdar:
1. Refactored the code around spring transaction semantics. Earlier we were using both - @Transactional and the TransactionTemplate.
2. Modified the MetacatInitializationService to  start up on ContextRefreshEvent instead of being explicitly called from the constructor. This way it will start once all the spring context is initialized.


metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlLookupService.java
```diff
@@ -101,7 +100,7 @@ public Lookup get(final String name) {
                 });
         } catch (EmptyResultDataAccessException e) {
             return null;
-        } catch (DataAccessException e) {
+        } catch (Exception e) {
```
zhljen: What's the exception here? If only the DataAccessException, should we be specific?

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlLookupService.java
```diff
@@ -173,53 +172,48 @@ public String getValue(final String name) {
...
-                }
-            });
-        } catch (CannotCreateTransactionException | DataAccessException e) {
+            Lookup lookup = null;
+            try {
+                lookup = findOrCreateLookupByName(name);
+            } catch (SQLException e) {
+                throw new CannotCreateTransactionException(
+                    "findOrCreateTagItemByName failed " + name, e);
+            }
```
zhljen: Previously, the doInTransaction clause is used and throws this when the exception happens inside. Since that explicit transaction is removed, we can throw or propagate the SQLException instead of throw a new CannotCreateTransactionException.

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlLookupService.java
```diff
@@ -268,35 +252,29 @@ public PreparedStatement createPreparedStatement(final Connection connection)
...
-                        throw new CannotCreateTransactionException(
-                            "findOrCreateTagItemByName failed " + name, e);
-                    }
+            Lookup lookup = null;
+            try {
+                lookup = findOrCreateLookupByName(name);
+            } catch (SQLException e) {
+                throw new CannotCreateTransactionException(
+                    "findOrCreateTagItemByName failed " + name, e);
+            }
```
zhljen: Similar as above.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorFastPartitionService.java
```diff
@@ -565,21 +568,12 @@ private Void populateParameters(final List<Long> ids,
                 final Long id = rs.getLong(idName);
                 final String key = rs.getString("param_key");
                 final String value = rs.getString("param_value");
-                Map<String, String> parameters = result.get(id);
-                if (parameters == null) {
-                    parameters = Maps.newHashMap();
-                    result.put(id, parameters);
-                }
+                final Map<String, String> parameters = result.computeIfAbsent(id, k -> Maps.newHashMap());
                 parameters.put(key, value);
```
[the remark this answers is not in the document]
ajoymajumdar: If key with id does not exist or is null, it will create the Map and put it.

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlLookupService.java
ajoymajumdar: We can be specific if we can differentiate between known exceptions vs unknown system exceptions. For now, we want to catch all and throw a UserMetadataServiceException.

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlServiceUtil.java
```diff
@@ -51,18 +43,16 @@ private MySqlServiceUtil() {
...
-     * @param jdbcUtil jdbc util
+     * @param jdbcTemplate jdbc template
      * @param sql      query sql
      * @param item     identifier
      * @return list of results
-     * @throws DataAccessException data access exception
      */
-    @Transactional(readOnly = true)
-    public static Set<String> getValues(final JdbcUtil jdbcUtil,
+    public static Set<String> getValues(final JdbcTemplate jdbcTemplate,
```
zhljen: It looks that Spring declarative transaction [does not support static method](https://stackoverflow.com/questions/23569732/transactional-with-static-method). I am not sure if this utility class will work here. You might want to test it out.

build.gradle
```diff
@@ -151,6 +151,7 @@ configure(javaProjects) {
         all*.exclude module: "jcl-over-slf4j"
         all*.exclude module: "log4j-over-slf4j"
         all*.exclude module: "slf4j-jdk14"
+        all*.exclude group: "ch.qos.logback", module: "logback-classic"
```
tgianos: if you're taking logback out what does the default logger use now? log4j?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorFastPartitionService.java
```diff
@@ -75,7 +76,7 @@
  * @since 1.0.0
  */
 @Slf4j
-@Transactional(readOnly = true)
+@Transactional("hiveTxManager")
```
tgianos: If all the methods in this class are readOnly why not leave the readOnly at the class level and remove the redundancy at the method level? Are you just concerned if for some reason a method is ever introduced that wasn't read only and so you're opting methods into readOnly mode instead of other way around?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorFastPartitionService.java
```diff
@@ -135,13 +136,13 @@ public HiveConnectorFastPartitionService(
         final HiveConnectorInfoConverter hiveMetacatConverters,
         final ConnectorContext connectorContext,
         final ThreadServiceManager threadServiceManager,
-        final DataSource dataSource,
+        @Qualifier("hiveJdbcTemplate") final JdbcTemplate jdbcTemplate,
```
tgianos: It doesn't look like this class is marked as a `Component` or `Service` so isn't this qualifier kind of useless? It should be where the bean is created (probably a configuration file) on the method input there

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorFastTableService.java
```diff
@@ -74,12 +74,12 @@ public HiveConnectorFastTableService(
         final HiveConnectorDatabaseService hiveConnectorDatabaseService,
         final HiveConnectorInfoConverter hiveMetacatConverters,
         final ConnectorContext connectorContext,
-        final DataSource dataSource,
+        @Qualifier("hiveJdbcTemplate") final JdbcTemplate jdbcTemplate,
```
tgianos: I might be missing something but the `@Autowired` here is kind of useless so is the `@Qualifier` cause the class isn't a classpath scanned bean...?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/configs/HiveConnectorFastServiceConfig.java
```diff
@@ -72,7 +71,7 @@ public HiveConnectorPartitionService fastHivePartitionService(
         final HiveConnectorInfoConverter hiveMetacatConverter,
         final ThreadServiceManager threadServiceManager,
         final ConnectorContext connectorContext,
-        @Qualifier("hiveDataSource") final DataSource dataSource,
+        @Qualifier("hiveJdbcTemplate") final JdbcTemplate hiveJdbcTemplate,
```
tgianos: This is the correct place for the `@Qualifier` the other one is redundant noise and ties the class to spring unnecessarily

metacat-main/src/test/groovy/com/netflix/metacat/main/services/MetacatInitializationServiceSpec.groovy
```diff
@@ -47,6 +47,7 @@ class MetacatInitializationServiceSpec extends Specification {
             threadServiceManager,
             thriftService
         )
+        initializationService.start(null)
```
tgianos: Null? What if you ever do something with the event in the Event listener. Right now you're serializing it as a string but if you ever change it to do something like `event.getSource()` this test is going to barf.

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlLookupService.java
```diff
@@ -61,17 +60,17 @@
...
 
     /**
      * Constructor.
      *
      * @param config     config
-     * @param dataSource datasource to user
+     * @param jdbcTemplate jdbc template
      */
-    public MySqlLookupService(final Config config, final DataSource dataSource) {
+    public MySqlLookupService(final Config config, @Qualifier("metadataJdbcTemplate") final JdbcTemplate jdbcTemplate) {
```
tgianos: Unnecessary `@Qualifier`

metacat-metadata-mysql/src/test/groovy/com/netflix/metacat/metadata/mysql/BaseSpec.groovy
```diff
@@ -34,13 +45,32 @@ import java.util.concurrent.atomic.AtomicBoolean
 import static java.lang.String.format
 
 @Ignore
```
tgianos: Does this still mean to be `@Ignore`?

build.gradle
ajoymajumdar: Yes, I thought it was always log4j.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorFastPartitionService.java
ajoymajumdar: Yes. Explicitly set readOnly at the method level if the method only does a read.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorFastPartitionService.java
ajoymajumdar: Agreed, this is redundant. It is already been defined in the config.

metacat-main/src/test/groovy/com/netflix/metacat/main/services/MetacatInitializationServiceSpec.groovy
ajoymajumdar: At present, we do not. Once we do, the test will raise a flag by failing.

metacat-metadata-mysql/src/test/groovy/com/netflix/metacat/metadata/mysql/BaseSpec.groovy
ajoymajumdar: This was added to not run the methods of this class.

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlServiceUtil.java
zhljen: Are we still using this one?

PR #191, opened 2017-10-06
Bypassing hive metastore to add/alter/drop partitions.

ajoymajumdar:
1. Implemented direct sqls to the metastore RDS for add/alter/drop partitions.
2. Added functional tests

PR #193, opened 2017-10-16
Fixed issue introduced in PR #191.

ajoymajumdar:
Issue: Partition parameters were not getting stored.


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlSavePartition.java
```diff
@@ -135,6 +135,13 @@ private void _insert(final QualifiedName tableQName, final Table table, final Ta
             for (int i = 0; i < partValues.size(); i++) {
                 partitionKeyValsValues.add(new Object[] {partId, partValues.get(i), i});
             }
+            // Partition parameters
+            final Map<String, String> parameters = partition.getMetadata();
+            if (parameters != null) {
```
tgianos: more of an all encompassing suggestion but we should start using `Optional` to reflect nullable condition as an API/module delineation level. I doubt you'll want to change this now but I think it makes for more manageable code rather than guessing or having to always check documentation to see if something is null.

PR #195, opened 2017-10-18
Added to get a counter of non-existing partition locations when creating new partitions.

ajoymajumdar:


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/metastore/MetacatHMSHandler.java
```diff
@@ -58,8 +58,12 @@
  * @since 1.0.0
  */
 public class MetacatHMSHandler extends HiveMetaStore.HMSHandler implements IMetacatHMSHandler {
+    private static final String REGISTRY_ID_IS_NOT_DIR = "metacat.hive.counter.partitionPathIsNotDir";
+
```
zhljen: This should belong to the HiveMetrics class.

PR #196, opened 2017-10-21
Fixed issue with SQL save partitions where trailing '/' was not getting removed from the partition uri.

ajoymajumdar:
In Hive, partition uri with trailing '/' is removed for new partitions but for altered partitions, it does not. With the new SQL save partitions logic, trailing '/' is removed in both cases.

### DIFF
--- a/metacat-common/src/main/java/com/netflix/metacat/common/MetacatRequestContext.java
+++ b/metacat-common/src/main/java/com/netflix/metacat/common/MetacatRequestContext.java
@@ -108,6 +108,22 @@ public class MetacatRequestContext {
         this.scheme = scheme;
     }
 
+    @Override
+    public String toString() {
+        final StringBuilder sb = new StringBuilder("MetacatRequestContext{");
+        sb.append("id='").append(id).append('\'');
+        sb.append(", timestamp=").append(timestamp);
+        sb.append(", userName='").append(userName).append('\'');
+        sb.append(", clientAppName='").append(clientAppName).append('\'');
+        sb.append(", clientId='").append(clientId).append('\'');
+        sb.append(", jobId='").append(jobId).append('\'');
+        sb.append(", dataTypeContext='").append(dataTypeContext).append('\'');
+        sb.append(", apiUri='").append(apiUri).append('\'');
+        sb.append(", scheme='").append(scheme).append('\'');
+        sb.append('}');
+        return sb.toString();
+    }
+
     /**
      * builder class for MetacatRequestContext.
      * @return the builder class for MetacatRequestContext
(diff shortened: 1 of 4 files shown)


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastPartitionService.java
```diff
@@ -177,18 +178,27 @@ private void createLocationForPartition(final QualifiedName tableQName,
...
+                path = warehouse.getDnsPath(new Path(location));
+            } catch (Exception e) {
+                throw new InvalidMetaException(String.format("Failed forming partition location; %s", location), e);
+            }
+        }
+        if (path != null) {
+            location = path.toString();
+            partitionInfo.getSerde().setUri(location);
+            if (doFileSystemCalls) {
+                try {
```
zhljen: We should add a metric here to collect how many filey system op are called.

metacat-common/src/main/java/com/netflix/metacat/common/MetacatRequestContext.java
```diff
@@ -108,6 +108,22 @@ protected MetacatRequestContext(
         this.scheme = scheme;
     }
 
+    @Override
+    public String toString() {
+        final StringBuilder sb = new StringBuilder("MetacatRequestContext{");
```
tgianos: Any particular reason you don't just use Jackson here and create a JSON object instead of all this manual serialization?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastPartitionService.java
```diff
@@ -169,6 +169,7 @@ private void createLocationForPartitions(final QualifiedName tableQName,
     private void createLocationForPartition(final QualifiedName tableQName,
         final PartitionInfo partitionInfo, final Table table, final boolean doFileSystemCalls) {
         String location = partitionInfo.getSerde().getUri();
+        Path path = null;
```
tgianos: Shouldn't this be final path? and is there any way through the if/else chain that doesn't result in an exception being thrown that this condition could even happen?

metacat-common/src/main/java/com/netflix/metacat/common/MetacatRequestContext.java
ajoymajumdar: Why would you want to create a JSON object and then do a toString.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastPartitionService.java
ajoymajumdar: Any runtime exception will end up failing the API call. We catch the checked exceptions and throw a relevant Metacat exception.

metacat-common/src/main/java/com/netflix/metacat/common/MetacatRequestContext.java
tgianos: I don't know the use of your toString here but I'm saying return a JSON string from the toString.

ajoymajumdar: It is a toString method that prints out the instance variable values. Today we log this for every API call. It is not in JSON format. I auto-generated this method in intellij.

PR #202, opened 2017-11-09
Added method to clean up obsolete definition metadata and tags.

ajoymajumdar:


metacat-main/src/main/java/com/netflix/metacat/main/api/v1/MetadataController.java
```diff
@@ -258,7 +238,24 @@ public void deleteDefinitionMetadata(
      */
     @RequestMapping(method = RequestMethod.DELETE, path = "/data/process")
```
zhljen: Should we name it as path = "/datametadata/process"? I am not sure about the 'process', but we should explicitly point out the datametadata and definitionmeatdata part.

metacat-main/src/main/java/com/netflix/metacat/main/services/MetadataService.java
```diff
@@ -37,15 +42,39 @@
  * Created by amajumdar on 9/26/16.
  */
```
zhljen: Can these follow the header format?

metacat-main/src/main/java/com/netflix/metacat/main/services/MetadataService.java
```diff
@@ -37,15 +42,39 @@
...
+        this.userMetadataService = userMetadataService;
+        this.tagService = tagService;
+        this.helper = helper;
+        this.registry = registry;
+    }
+
+    /**
      * Deletes all the data metadata marked for deletion.
      */
     public void processDeletedDataMetadata() {
```
zhljen: I am not sure if this function is trying to delete the data metadata or process the deleted data. The comments and the function name confused me.

metacat-main/src/main/java/com/netflix/metacat/main/services/MetadataService.java
```diff
@@ -102,4 +131,73 @@ public void processDeletedDataMetadata() {
...
+        final ListeningExecutorService service = threadServiceManager.getExecutor();
+        while (offset == 0 || dtos.size() == limit) {
+            dtos = userMetadataService.searchDefinitionMetadatas(null, null, null,
+                "id", null, offset, limit);
+            dtos.forEach(dto ->
+                service.submit(() -> deleteDefinitionMetadata(dto.getName(), false, metacatRequestContext)));
+            offset += limit;
+        }
+        threadServiceManager.stop();
+        log.info("End deleting obsolete definition metadata");
```
zhljen: We should log out the number of deleted records. It will be helpful to bucket them into catalogs, but at least the count will be useful.

metacat-main/src/main/java/com/netflix/metacat/main/services/MetadataService.java
ajoymajumdar: It is deleting obsolete data metadata. I agree the method name is confusing. This was written long back.

metacat-main/src/main/java/com/netflix/metacat/main/api/v1/MetadataController.java
ajoymajumdar: This is a metadata controller. Hence I named it data or definition.

zhljen: The ending point ( path ) should be self-explained, the client shouldn't know what controller it connects to.

zhljen: How about rename to path = "/obsoleteData" and "/obsoleteDefinition"? I do want to make the ending point self-explained what will actually happen.  The method name 'Delete' is explained the action, but the path /data/process is not clear.

ajoymajumdar: I am not good with names but I will rename it to /data/cleanup and /definition/cleanup. Is it fine?

zhljen: I thought about the 'cleanup' too, but it reads like 'delete /data/cleanup', and recommended 'delete obsoletedata'.  Both of them are better than 'process'..

tgianos: What @zhljen is saying here is really more of a RPC vs. REST argument. I see what she's saying about making it clear but I think it breaks consistency with a lot of our API.

If it isn't clear via the RESTful path of a `DELETE` to `/data/process` then that is the fault of the naming of those payloads and we probably should try to fix that root cause rather than hack some RPC method paths into it to get around it.

just my 2 cents.

ajoymajumdar: This is not the DELETE for the resource else it would have been /definition or /data with DELETE method. This is not a resource API. It is an API for cleaning up the invalid entries in definition_metadata and data_metadata and hence I named it /process but we could be more precise by saying cleanup..

tgianos: Lets talk about it on Monday in the meeting to come to a good answer for this going forward. @zhljen and I were talking about it in the office and we probably just need to come up with a consistent answer going forward. This seems to be a administration type API and perhaps we need to break these kinds of things apart

zhljen:
We can go like for now since the deletion of the obsolete data is need immediately. But I agree with Tom that we need a cleaner way to organize/implement these admin APIs.

PR #203, opened 2017-11-20
Sharded table 'definition_metadata' into two tables: 'definition_metadata' and 'partition_deifnition_metadata'.

ajoymajumdar:
'partition_definition_metadata' stores metadata about the partitions. Metadata about catalog, database and tables are stored in 'definition_metadata'.

### DIFF
--- a/metacat-functional-tests/metacat-test-cluster/datastores/mysql/docker-entrypoint-initdb.d/metacat.sql
+++ b/metacat-functional-tests/metacat-test-cluster/datastores/mysql/docker-entrypoint-initdb.d/metacat.sql
@@ -65,6 +65,24 @@ CREATE TABLE `definition_metadata` (
 ) ENGINE=InnoDB AUTO_INCREMENT=10078235 DEFAULT CHARSET=latin1;
 /*!40101 SET character_set_client = @saved_cs_client */;
 
+DROP TABLE IF EXISTS `partition_definition_metadata`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `partition_definition_metadata` (
+  `id` bigint(20) NOT NULL AUTO_INCREMENT,
+  `version` bigint(20) NOT NULL,
+  `created_by` varchar(255) NOT NULL,
+  `data` longtext NOT NULL,
+  `date_created` datetime NOT NULL,
+  `last_updated` datetime NOT NULL,
+  `last_updated_by` varchar(255) NOT NULL,
+  `name` varchar(255) NOT NULL,
+  PRIMARY KEY (`id`),
+  UNIQUE KEY `name` (`name`)
+) ENGINE=InnoDB AUTO_INCREMENT=10078235 DEFAULT CHARSET=latin1;
+/*!40101 SET character_set_client = @saved_cs_client */;
+
+
 --
 -- Table structure for table `lookup`
 --
--- a/metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MysqlUserMetadataService.java
+++ b/metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MysqlUserMetadataService.java
@@ -48,6 +48,13 @@ import java.util.stream.Collectors;
 
 /**
  * User metadata service.
+ *
+ * Definition metadata (business metadata about the logical schema definition) is stored in two tables. Definition
+ * metadata about the partitions are stored in 'partition_definition_metadata' table. Definition metadata about the
+ * catalogs, databases and tables are stored in 'definition_metadata' table.
+ *
+ * Data metadata (metadata about the data stored in the location referred by the schema). This information is stored in
+ * 'data_metadata' table.
  */
 @Slf4j
 @SuppressFBWarnings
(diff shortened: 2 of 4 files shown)


metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MysqlUserMetadataService.java
```diff
@@ -191,12 +190,25 @@ private void _deleteDefinitionMetadatas(
...
-            jdbcTemplate.update(
-                String.format(SQL.DELETE_DEFINITION_METADATA, Joiner.on(",").skipNulls().join(paramVariables)),
-                (Object[]) aNames);
+            final SqlParameterValue[] aPartitionNames = names.stream().filter(QualifiedName::isPartitionDefinition)
+                .map(n -> new SqlParameterValue(Types.VARCHAR, n))
+                .toArray(SqlParameterValue[]::new);
+            if (aNames.length > 0) {
+                final List<String> paramVariables = Arrays.stream(aNames).map(s -> "?").collect(Collectors.toList());
+                jdbcTemplate.update(
+                    String.format(SQL.DELETE_DEFINITION_METADATA, Joiner.on(",").skipNulls().join(paramVariables)),
```
zhljen: Is this table definition metadata only or including database metadata? If just table only, we should name it explicitly as table definition metadata.

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MysqlUserMetadataService.java
```diff
@@ -331,16 +345,29 @@ private void _deleteDataMetadatas(
     @Transactional(readOnly = true)
     public Map<String, ObjectNode> getDefinitionMetadataMap(
         @Nonnull final List<QualifiedName> names) {
-        if (!names.isEmpty()) {
-            final List<List<QualifiedName>> parts = Lists.partition(names, config.getUserMetadataMaxInClauseItems());
-            return parts.stream()
+        final List<QualifiedName> oNames = names.stream().filter(name -> !name.isPartitionDefinition()).collect(
```
zhljen: Does this function returns both the table definition metadata map and the partition metadata map? If we don't want to change the API, we should at least separate these two parts as two functions internally for clarity.

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MysqlUserMetadataService.java
```diff
@@ -605,6 +635,16 @@ public void saveMetadatas(final String user, final List<? extends HasMetadata> m
                         jdbcTemplate.batchUpdate(SQL.UPDATE_DEFINITION_METADATA, updateDefinitionMetadatas,
                             new int[]{Types.VARCHAR, Types.VARCHAR, Types.VARCHAR});
                     }
+                    if (!insertPartitionDefinitionMetadatas.isEmpty()) {
```
zhljen: Similar as above, we should regroup the partition metadata insert into one function for clarity, or it can be used to update/insert partition definition metadata only.

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MysqlUserMetadataService.java
ajoymajumdar: Mentioned in the description. Will add it in the comments in the class.

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MysqlUserMetadataService.java
ajoymajumdar: Yes, it does. Since sharding the table is an internal change for optimization purpose, we do not have a requirement to expose an API to separate them.For clarity,  I will create internal private methods.

metacat-functional-tests/metacat-test-cluster/datastores/mysql/docker-entrypoint-initdb.d/metacat.sql
```diff
@@ -65,6 +65,24 @@ CREATE TABLE `definition_metadata` (
...
+  `version` bigint(20) NOT NULL,
+  `created_by` varchar(255) NOT NULL,
+  `data` longtext NOT NULL,
+  `date_created` datetime NOT NULL,
+  `last_updated` datetime NOT NULL,
+  `last_updated_by` varchar(255) NOT NULL,
+  `name` varchar(255) NOT NULL,
+  PRIMARY KEY (`id`),
+  UNIQUE KEY `name` (`name`)
+) ENGINE=InnoDB AUTO_INCREMENT=10078235 DEFAULT CHARSET=latin1;
```
tgianos: Why are you setting the auto increment number to start with?

metacat-metadata-mysql/src/test/resources/sql/prepare-test.sql
```diff
@@ -36,6 +36,20 @@ CREATE TABLE metacat.definition_metadata (
   UNIQUE KEY name (name)
 ) DEFAULT CHARSET=latin1;
 
+-- Create syntax for TABLE 'definition_metadata'
+CREATE TABLE metacat.partition_definition_metadata (
+  id bigint(20) NOT NULL AUTO_INCREMENT,
+  version bigint(20) NOT NULL,
+  created_by varchar(255) NOT NULL,
+  data longtext NOT NULL,
```
tgianos: This is a huge data type. you really need this?

metacat-metadata-mysql/src/test/resources/sql/prepare-test.sql
```diff
@@ -36,6 +36,20 @@ CREATE TABLE metacat.definition_metadata (
...
+  date_created datetime NOT NULL,
```
tgianos: if you want greater granularity you can set `datetime(3)` for milliseconds... not sure if you want that or not but that's what is done in Genie

metacat-functional-tests/metacat-test-cluster/datastores/mysql/docker-entrypoint-initdb.d/metacat.sql
ajoymajumdar: Just a number to start with. Copied the schema from our current database.

metacat-metadata-mysql/src/test/resources/sql/prepare-test.sql
ajoymajumdar: Yes, we are using this since the JSON string can be huge.

PR #206, opened 2017-11-27
Modified API getDatabase to return the list of tables based on a request boolean param 'includeTableNames'.

ajoymajumdar:


metacat-thrift/src/main/java/com/netflix/metacat/thrift/CatalogThriftHiveMetastore.java
```diff
@@ -841,7 +841,7 @@ public CurrentNotificationEventId get_current_notificationEventId() throws TExce
     public Database get_database(final String name) throws TException {
         return requestWrapper("get_database", new Object[]{name}, () -> {
             final String databaseName = normalizeIdentifier(name);
-            final DatabaseDto dto = v1.getDatabase(catalogName, databaseName, true);
+            final DatabaseDto dto = v1.getDatabase(catalogName, databaseName, true, false);
```
zhljen: We stopped return the table names from get_database() call. Will this break anything?

metacat-client/src/main/java/com/netflix/metacat/client/api/MetacatV1.java
```diff
@@ -236,7 +237,10 @@ DatabaseDto getDatabase(
             String databaseName,
         @DefaultValue("true")
         @QueryParam("includeUserMetadata")
-            Boolean includeUserMetadata
+            Boolean includeUserMetadata,
+        @DefaultValue("true")
+        @QueryParam("includeTableNames")
+            Boolean includeTableNames
```
tgianos: Since you're using a default value do you still need to autobox it into a `Boolean`? Can you just use the native boolean type and save the small bit over overhead?

metacat-main/src/main/java/com/netflix/metacat/main/services/DatabaseService.java
```diff
@@ -24,7 +24,8 @@
      * Gets the database for the given name.
      * @param name name
      * @param includeUserMetadata if true, will also include the user metadata
+     * @param includeTableNames   if true, include the list of table names
      * @return database info
      */
-    DatabaseDto get(QualifiedName name, boolean includeUserMetadata);
+    DatabaseDto get(QualifiedName name, boolean includeUserMetadata, boolean includeTableNames);
```
tgianos: Since this seems to be true most places would it be cleaner just to make a separate interface `getWithoutTableNames` to make it very clear? The inside the implementation you can have a private method with the boolean switch

metacat-thrift/src/main/java/com/netflix/metacat/thrift/CatalogThriftHiveMetastore.java
ajoymajumdar: In this method, the list of table names is not used.

metacat-client/src/main/java/com/netflix/metacat/client/api/MetacatV1.java
ajoymajumdar: I kept it consistent with other REST APIs.

metacat-main/src/main/java/com/netflix/metacat/main/services/DatabaseService.java
ajoymajumdar: This looks much simpler and inline with 'includeUserMetadata' flag than adding another method.

PR #207, opened 2017-11-28
Optimize delete hive partitions by using direct sql.

ajoymajumdar:


metacat-functional-tests/src/functionalTest/groovy/com/netflix/metacat/MetacatSmokeSpec.groovy
```diff
@@ -821,10 +821,6 @@ class MetacatSmokeSpec extends Specification {
         'embedded-hive-metastore'       | 'smoke_db'   | 'part'    | ['one=test', 'one=invalid']
         'embedded-hive-metastore'       | 'smoke_db'   | 'part'    | ['one=test', 'one=invalid']
         'embedded-hive-metastore'       | 'smoke_db'   | 'invalid' | ['one=test', 'one=invalid']
-        'embedded-fast-hive-metastore'  | 'fsmoke_db'  | 'part'    | ['one=invalid']
```
zhljen: The sql deletion doesn't throw exception if the partition is invalid and I am not sure if this will break anything. Considering the intention of the change is to optimize the performance as we can, it's ok to remove these tests here. But we should put more explicit explanation in the codes in case issues come out later.

PR #219, opened 2018-01-17
Iceberg table update

ajoymajumdar:
Initial changes for iceberg support. On iceberg table update, the table is locked and validated for check/put. If valid, it updates the table. If not valid, it will throw a PreconditionFailedException.

### DIFF
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/exception/TablePreconditionFailedException.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/exception/TablePreconditionFailedException.java
@@ -1,0 +1,39 @@
+/*
+ *
+ *  Copyright 2016 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.metacat.common.server.connectors.exception;
+
+import com.netflix.metacat.common.QualifiedName;
+
+import javax.annotation.Nullable;
+
+/**
+ * Exception when table is not found.
+ *
+ * @author amajumdar
+ */
+public class TablePreconditionFailedException extends ConnectorException {
+    /**
+     * Constructor.
+     *
+     * @param name  qualified name of the table
+     * @param message error cause
+     */
+    public TablePreconditionFailedException(final QualifiedName name, @Nullable final String message) {
+        super(String.format("Precondition failed to update table %s. %s", name, message));
+    }
+}
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/exception/TablePreconditionFailedException.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/exception/TablePreconditionFailedException.java
@@ -1,6 +1,6 @@
 /*
  *
- *  Copyright 2016 Netflix, Inc.
+ *  Copyright 2018 Netflix, Inc.
  *
  *     Licensed under the Apache License, Version 2.0 (the "License");
  *     you may not use this file except in compliance with the License.
--- a/metacat-common/src/main/java/com/netflix/metacat/common/exception/MetacatPreconditionFailedException.java
+++ b/metacat-common/src/main/java/com/netflix/metacat/common/exception/MetacatPreconditionFailedException.java
@@ -1,6 +1,6 @@
 /*
  *
- *  Copyright 2016 Netflix, Inc.
+ *  Copyright 2018 Netflix, Inc.
  *
  *     Licensed under the Apache License, Version 2.0 (the "License");
  *     you may not use this file except in compliance with the License.
@@ -20,6 +20,7 @@ package com.netflix.metacat.common.exception;
 /**
  * Exception when operation failed precondition. Ex. if a locked table is updated, this exception will be thrown.
  *
+ * @since 1.2.0
  * @author amajumdar
  */
 public class MetacatPreconditionFailedException extends MetacatException {
--- a/metacat-common/src/main/java/com/netflix/metacat/common/exception/MetacatPreconditionFailedException.java
+++ b/metacat-common/src/main/java/com/netflix/metacat/common/exception/MetacatPreconditionFailedException.java
@@ -1,0 +1,34 @@
+/*
+ *
+ *  Copyright 2016 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.metacat.common.exception;
+
+/**
+ * Exception when operation failed precondition. Ex. if a locked table is updated, this exception will be thrown.
+ *
+ * @author amajumdar
+ */
+public class MetacatPreconditionFailedException extends MetacatException {
+    /**
+     * Constructor.
+     *
+     * @param reason exception message
+     */
+    public MetacatPreconditionFailedException(final String reason) {
+        super(reason);
+    }
+}
--- a/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/configs/HiveConnectorFastServiceConfig.java
+++ b/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/configs/HiveConnectorFastServiceConfig.java
@@ -16,14 +16,15 @@ package com.netflix.metacat.connector.hive.configs;
 import com.netflix.metacat.common.server.connectors.ConnectorContext;
 import com.netflix.metacat.common.server.util.ThreadServiceManager;
 import com.netflix.metacat.connector.hive.HiveConnectorDatabaseService;
-import com.netflix.metacat.connector.hive.sql.DirectSqlGetPartition;
-import com.netflix.metacat.connector.hive.sql.DirectSqlSavePartition;
-import com.netflix.metacat.connector.hive.sql.HiveConnectorFastPartitionService;
-import com.netflix.metacat.connector.hive.sql.HiveConnectorFastTableService;
 import com.netflix.metacat.connector.hive.HiveConnectorPartitionService;
 import com.netflix.metacat.connector.hive.HiveConnectorTableService;
 import com.netflix.metacat.connector.hive.IMetacatHiveClient;
 import com.netflix.metacat.connector.hive.converters.HiveConnectorInfoConverter;
+import com.netflix.metacat.connector.hive.sql.DirectSqlGetPartition;
+import com.netflix.metacat.connector.hive.sql.DirectSqlSavePartition;
+import com.netflix.metacat.connector.hive.sql.DirectSqlTable;
+import com.netflix.metacat.connector.hive.sql.HiveConnectorFastPartitionService;
+import com.netflix.metacat.connector.hive.sql.HiveConnectorFastTableService;
 import com.netflix.metacat.connector.hive.sql.SequenceGeneration;
 import com.netflix.metacat.connector.hive.util.HiveConnectorFastServiceMetric;
 import org.apache.hadoop.hive.metastore.Warehouse;
@@ -152,14 +153,34 @@ public class HiveConnectorFastServiceConfig {
     }
 
     /**
+     * Service to get partitions.
+     *
+     * @param connectorContext     connector config
+     * @param hiveJdbcTemplate     hive JDBC template
+     * @param serviceMetric        fast service metric
+     * @return HiveConnectorPartitionService
+     */
+    @Bean
+    public DirectSqlTable directSqlTable(
+        final ConnectorContext connectorContext,
+        @Qualifier("hiveJdbcTemplate") final JdbcTemplate hiveJdbcTemplate,
+        final HiveConnectorFastServiceMetric serviceMetric
+    ) {
+        return new DirectSqlTable(
+            connectorContext,
+            hiveJdbcTemplate,
+            serviceMetric
+        );
+    }
+
+    /**
      * create hive connector fast table service.
      *
      * @param metacatHiveClient            metacat hive client
      * @param hiveMetacatConverters        hive metacat converters
      * @param hiveConnectorDatabaseService hive database service
      * @param connectorContext             server context
-     * @param hiveJdbcTemplate             hive JDBC template
-     * @param serviceMetric                fast service metric
+     * @param directSqlTable               table jpa service
      * @return HiveConnectorFastTableService
      */
     @Bean
@@ -168,8 +189,7 @@ public class HiveConnectorFastServiceConfig {
         final HiveConnectorInfoConverter hiveMetacatConverters,
         final HiveConnectorDatabaseService hiveConnectorDatabaseService,
         final ConnectorContext connectorContext,
-        @Qualifier("hiveJdbcTemplate") final JdbcTemplate hiveJdbcTemplate,
-        final HiveConnectorFastServiceMetric serviceMetric
+        final DirectSqlTable directSqlTable
     ) {
         return new HiveConnectorFastTableService(
             connectorContext.getCatalogName(),
@@ -177,8 +197,7 @@ public class HiveConnectorFastServiceConfig {
             hiveConnectorDatabaseService,
             hiveMetacatConverters,
             connectorContext,
-            hiveJdbcTemplate,
-            serviceMetric
+            directSqlTable
         );
     }
 }
--- a/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/configs/HiveConnectorFastServiceConfig.java
+++ b/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/configs/HiveConnectorFastServiceConfig.java
@@ -153,7 +153,7 @@ public class HiveConnectorFastServiceConfig {
     }
 
     /**
-     * Service to get partitions.
+     * Data access service to get partitions.
      *
      * @param connectorContext     connector config
      * @param hiveJdbcTemplate     hive JDBC template
--- a/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
+++ b/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
@@ -1,0 +1,217 @@
+/*
+ *
+ *  Copyright 2016 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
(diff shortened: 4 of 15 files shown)


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
```diff
@@ -0,0 +1,217 @@
...
+
+    /**
+     * Locks the iceberg table for update so that no other request can modify the table.
+     * @param tableId table internal id
+     */
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    public void lockIcebergTable(final Long tableId) {
+        Boolean isLocked = null;
+        try {
+            isLocked = jdbcTemplate.queryForObject(SQL.ICEBERG_TABLE_LOCK,
```
zhljen: The sql here is to lock a table parameter row, which is quite generic. The sql shall reflect what this is doing. So instead of saying ICEBERG_TABLE_LOCK, we can say TABLE_PARAM_LOCK. We can still leave the function name as lockIcebergTable to be specific and explicit.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
```diff
@@ -0,0 +1,217 @@
...
+    private void updateTableParam(final Long tableId, final String parameter, final int value) {
+        jdbcTemplate.update(SQL.UPDATE_TABLE_PARAMS, new SqlParameterValue(Types.VARCHAR, value),
+            new SqlParameterValue(Types.BIGINT, tableId), new SqlParameterValue(Types.VARCHAR, parameter));
+    }
+
+    /**
+     * Returns the table internal id.
+     * @param tableName table name
+     * @return table id
+     */
```
zhljen: Should this include @Transactional(readOnly = true)?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastTableService.java
```diff
@@ -114,43 +89,59 @@ public boolean exists(final ConnectorRequestContext requestContext, final Qualif
...
+            final Long tableId = directSqlTable.getTableId(tableName);
+            try {
+                directSqlTable.lockIcebergTable(tableId);
+                try {
+                    final TableInfo existingTableInfo = get(requestContext, tableInfo.getName());
+                    if (isIcebergTable(existingTableInfo) && isValidIcebergUpdate(existingTableInfo, tableInfo)) {
+                        final Table existingTable = getHiveMetacatConverters().fromTableInfo(existingTableInfo);
+                        super.update(requestContext, existingTable, tableInfo);
+                    } else {
+                        throw new IllegalStateException("Invalid iceberg table metadata");
```
zhljen: This invalid metadata error should be separated from the lock failure, which is also IllegalStateException.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastTableService.java
```diff
@@ -114,43 +89,59 @@ public boolean exists(final ConnectorRequestContext requestContext, final Qualif
...
+                    }
+                } finally {
+                    directSqlTable.unlockIcebergTable(tableId);
+                }
+            } catch (IllegalStateException e) {
+                throw new TablePreconditionFailedException(tableName, e.getMessage());
+            }
```
zhljen: Since they are all illegalStateException, how does the client distinguish table-in-lock to retry or bad metadata request?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastTableService.java
```diff
@@ -114,43 +89,59 @@ public boolean exists(final ConnectorRequestContext requestContext, final Qualif
...
+    @Override
+    public void update(final ConnectorRequestContext requestContext, final TableInfo tableInfo) {
+        if (isIcebergTable(tableInfo)) {
+            final QualifiedName tableName = tableInfo.getName();
+            final Long tableId = directSqlTable.getTableId(tableName);
+            try {
+                directSqlTable.lockIcebergTable(tableId);
+                try {
+                    final TableInfo existingTableInfo = get(requestContext, tableInfo.getName());
+                    if (isIcebergTable(existingTableInfo) && isValidIcebergUpdate(existingTableInfo, tableInfo)) {
```
zhljen: Should ` isIcebergTable(existingTableInfo)` be moved to the top? If the existing table not an icebergTable, are we able to lock it? I guess, by inserting a location? But is this valid?  To valid a table to be iceberg should be stay in one function, maybe we can have one function to check both the new table info request and the existing table are both valid iceberg table.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
ajoymajumdar: yes.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
ajoymajumdar: Will change.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastTableService.java
ajoymajumdar: For iceberg operation it does not matter but we agreed to throw one exception stating 'preconditionfailed'.

tgianos:
You're putting this into a patch release? seems like a large feature?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/configs/HiveConnectorFastServiceConfig.java
```diff
@@ -151,15 +152,35 @@ public SequenceGeneration sequenceGeneration(
         return new SequenceGeneration(hiveJdbcTemplate);
     }
 
+    /**
+     * Service to get partitions.
```
tgianos: This is a service but the class name doesn't seem to reflect that at all?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
```diff
@@ -0,0 +1,217 @@
+/*
+ *
+ *  Copyright 2016 Netflix, Inc.
```
tgianos: As @zhljen pointed out all your dates seem wrong

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
```diff
@@ -0,0 +1,217 @@
...
+        boolean result = false;
+        try {
+            final Object qResult = jdbcTemplate.queryForObject(SQL.EXIST_TABLE_BY_NAME,
+                new String[]{name.getDatabaseName(), name.getTableName()},
+                new int[]{Types.VARCHAR, Types.VARCHAR}, Integer.class);
+            if (qResult != null) {
+                result = true;
+            }
+        } catch (EmptyResultDataAccessException e) {
+            return false;
```
tgianos: Not even a log statement that this occurred?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
```diff
@@ -0,0 +1,217 @@
...
+     * @param uris         locations
+     * @param prefixSearch if true, we look for tables whose location starts with the given <code>uri</code>
+     * @return map of uri to list of partition names
+     * @throws UnsupportedOperationException If the connector doesn't implement this method
+     */
+    @Transactional(readOnly = true)
+    public Map<String, List<QualifiedName>> getTableNames(final List<String> uris, final boolean prefixSearch) {
+        final long start = registry.clock().wallTime();
+        // Create the sql
+        final StringBuilder queryBuilder = new StringBuilder(SQL.GET_TABLE_NAMES_BY_URI);
```
tgianos: For testing you might want to think about breaking these SQL building blocks into their own method so you can quickly check them for bugs.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastTableService.java
ajoymajumdar: Client gets the PreConditionFailed exception with the detailed message if it is locked or with invalid metadata. We could use a error code if needed.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/exception/TablePreconditionFailedException.java
```diff
@@ -0,0 +1,39 @@
...
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.metacat.common.server.connectors.exception;
+
+import com.netflix.metacat.common.QualifiedName;
+
+import javax.annotation.Nullable;
+
```
zhljen: Think we want to move to NonNull as default. We may leave it as this for now.

metacat-common/src/main/java/com/netflix/metacat/common/exception/MetacatPreconditionFailedException.java
```diff
@@ -0,0 +1,34 @@
+/*
+ *
+ *  Copyright 2016 Netflix, Inc.
+ *
```
zhljen: This should be 2018

metacat-common/src/main/java/com/netflix/metacat/common/exception/MetacatPreconditionFailedException.java
```diff
@@ -0,0 +1,34 @@
...
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.metacat.common.exception;
+
+/**
+ * Exception when operation failed precondition. Ex. if a locked table is updated, this exception will be thrown.
+ *
+ * @author amajumdar
```
zhljen: To add `@since` annotation

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
```diff
@@ -0,0 +1,234 @@
...
+ * This class makes direct sql calls to get/set table metadata.
+ *
+ * @author amajumdar
+ * @since 1.1.1
+ */
+@Slf4j
+@Transactional("hiveTxManager")
+public class DirectSqlTable {
+    protected static final String PARAM_TABLE_TYPE = "table_type";
+    protected static final String PARAM_METADATA_LOCATION = "metadata_location";
```
zhljen: Can you please add comment to explain this constant? It's quite special and the name is not self-explained.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
```diff
@@ -0,0 +1,234 @@
...
+    protected static final String PARAM_PREVIOUS_METADATA_LOCATION = "previous_metadata_location";
+    protected static final String ICEBERG_TABLE_TYPE = "iceberg";
```
zhljen: Make 'iceberg' upper case (or ICEBERG_TABLE) to be consistent with the rest of the table type, 'EXTERNAL_TABLE', 'VIRTUAL_VIEW"...

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
```diff
@@ -0,0 +1,234 @@
...
+        try {
+            isLocked = jdbcTemplate.queryForObject(SQL.TABLE_PARAM_LOCK,
+                new SqlParameterValue[] {new SqlParameterValue(Types.BIGINT, tableId),
+                    new SqlParameterValue(Types.VARCHAR, PARAM_METADATA_LOCK), }, Boolean.class);
+        } catch (EmptyResultDataAccessException ignored) { }
+        if (isLocked == null) {
+            insertTableParam(tableId, PARAM_METADATA_LOCK, 1);
+        } else if (isLocked) {
+            throw new IllegalStateException("Iceberg table is locked.");
+        } else {
```
zhljen: We can include the tableName in the exception string, which can help us to trace later. I am not sure if this message will be printed out somewhere. In case there're multiple locking failures from our side, we need a way to figure out which table has locking problems. We can do atlas metrics too, but think logging it out is easier here.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastTableService.java
```diff
@@ -114,43 +84,67 @@ public boolean exists(final ConnectorRequestContext requestContext, final Qualif
...
+        if (isIcebergTable(tableInfo)) {
+            final QualifiedName tableName = tableInfo.getName();
+            final Long tableId = directSqlTable.getTableId(tableName);
+            try {
+                log.debug("Locking Iceberg table {}", tableName);
+                directSqlTable.lockIcebergTable(tableId, tableName);
+                try {
+                    final TableInfo existingTableInfo = get(requestContext, tableInfo.getName());
+                    if (!isValidIcebergUpdate(existingTableInfo, tableInfo)) {
+                        throw new IllegalStateException("Invalid iceberg table metadata.");
```
zhljen: We should mark this as invalidMetaDataException, and separate this from real IllegalStateException ( locking exception ).

zhljen: I think you mark it as IllegalStateException is because of it's part of the check-put condition check. This is different from the real metadata validation check.  We should put the explicit message in the exception - "the existing metadata location does not match".

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastTableService.java
```diff
@@ -114,43 +84,67 @@ public boolean exists(final ConnectorRequestContext requestContext, final Qualif
...
+        } else {
+            log.info("Invalid iceberg table metadata location (expected:{}, given:{})",
+                existingMetadataLocation, previousMetadataLocation);
+            return false;
         }
-        return result;
     }
 
+    private boolean isIcebergTable(final TableInfo tableInfo) {
+        return tableInfo.getMetadata() != null
```
zhljen: Looks to me that this can be used in the existing table check in lockingIcebergTable() function.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
ajoymajumdar: This is what is being used by iceberg. It should be fine.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastTableService.java
ajoymajumdar: what do u mean?

ajoymajumdar:
I did consider adding a switch to turn on the feature but if does not matter since it will apply only to tables that have the table_type as iceberg.

ajoymajumdar:
We do have functional tests

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
zhljen: It looks ugly, they haven't in prod yet and we can ask for change.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastTableService.java
zhljen: I mean the logic to check existing table. Is this function doing the same thing?
`try {
             tableType = jdbcTemplate.queryForObject(SQL.TABLE_PARAM_LOCK,
                 new SqlParameterValue[] {new SqlParameterValue(Types.BIGINT, tableId),
                     new SqlParameterValue(Types.VARCHAR, PARAM_TABLE_TYPE), }, String.class);
         } catch (EmptyResultDataAccessException ignored) { }
         if (tableType == null || !ICEBERG_TABLE_TYPE.equals(tableType)) {
             throw new InvalidMetaException(tableName, "Originally table is not of type iceberg", null);
         }`

metacat-functional-tests/src/functionalTest/groovy/com/netflix/metacat/MetacatSmokeSpec.groovy
```diff
@@ -261,6 +262,48 @@ class MetacatSmokeSpec extends Specification {
         'invalid-catalog'               | 'smoke_db1'  | 'z'                 | true   | false   | MetacatNotFoundException.class
     }
 
+    @Unroll
+    def "Test create/update iceberg table"() {
+        given:
```
zhljen: Why does the test stays in SmokeSpec instead of the functionalSpec? Feel majority of the functional tests are in the functionalSpec, and some of them are here. I missed it in the first place since I thought we should put the functionalSpec.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastTableService.java
ajoymajumdar: Yeah, we could. I want to keep it outside of the lock logic. I pushed the table_type check based on your comments.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastTableService.java
```diff
@@ -114,43 +84,68 @@ public boolean exists(final ConnectorRequestContext requestContext, final Qualif
...
+    private boolean isValidIcebergUpdate(final TableInfo existingTableInfo, final TableInfo newTableInfo) {
+        final Map<String, String> existingMetadata = existingTableInfo.getMetadata();
+        final Map<String, String> newMetadata = newTableInfo.getMetadata();
+        final String existingMetadataLocation = existingMetadata != null
+            ? existingMetadata.get(DirectSqlTable.PARAM_METADATA_LOCATION) : null;
+        final String previousMetadataLocation = newMetadata != null
+            ? newMetadata.get(DirectSqlTable.PARAM_PREVIOUS_METADATA_LOCATION) : null;
+        if (StringUtils.isNotBlank(existingMetadataLocation)
+            && Objects.equals(existingMetadataLocation, previousMetadataLocation)) {
+            return true;
```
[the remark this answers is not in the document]
ajoymajumdar: ok. Made the changes to be more precise.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastTableService.java
```diff
@@ -114,43 +84,71 @@ public boolean exists(final ConnectorRequestContext requestContext, final Qualif
...
+    }
+
+    private void validateIcebergUpdate(final TableInfo existingTableInfo, final TableInfo newTableInfo) {
+        final Map<String, String> existingMetadata = existingTableInfo.getMetadata();
+        final Map<String, String> newMetadata = newTableInfo.getMetadata();
+        final String existingMetadataLocation = existingMetadata != null
+            ? existingMetadata.get(DirectSqlTable.PARAM_METADATA_LOCATION) : null;
+        final String previousMetadataLocation = newMetadata != null
+            ? newMetadata.get(DirectSqlTable.PARAM_PREVIOUS_METADATA_LOCATION) : null;
+        if (StringUtils.isBlank(existingMetadataLocation)) {
```
zhljen: Thanks for updating. Feels the multiple if else can be better.

PR #220, opened 2018-01-22
Added support to not delete definition metadata when dropping a table.

ajoymajumdar:
Action to delete definition metadata now depends on a configuration. When a whitelist is defined, definition metadata for those tables should be deleted.

### DIFF
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/properties/Definition.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/properties/Definition.java
@@ -1,0 +1,73 @@
+/*
+ *
+ *  Copyright 2017 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.metacat.common.server.properties;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.netflix.metacat.common.QualifiedName;
+import lombok.NonNull;
+
+import java.util.Set;
+
+/**
+ * Definition metadata related properties.
+ *
+ * @author amajumdar
+ * @since 1.1.1
+ */
+@lombok.Data
+public class Definition {
+
+    @NonNull
+    private Metadata metadata = new Metadata();
+
+    /**
+     * Metadata related properties.
+     *
+     * @author amajumdar
+     * @since 1.1.1
+     */
+    @lombok.Data
+    public static class Metadata {
+
+        @NonNull
+        private Delete delete = new Delete();
+
+        /**
+         * Delete related properties.
+         *
+         * @author amajumdar
+         * @since 1.1.1
+         */
+        @lombok.Data
+        public static class Delete {
+            private boolean enableForTable = true;
+            private String enableDeleteForQualifiedNames;
+
+            /**
+             * Get the names stored in the variable as a List of fully qualified names.
+             *
+             * @return The names as a list or empty list if {@code enableDeleteForQualifiedNames} is null or empty
+             */
+            @JsonIgnore
+            public Set<QualifiedName> getQualifiedNamesEnabledForDelete() {
+                return PropertyUtils.delimitedStringsToQualifiedNamesSet(this.enableDeleteForQualifiedNames,
+                    ',');
+            }
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


metacat-common-server/src/main/java/com/netflix/metacat/common/server/properties/Definition.java
```diff
@@ -0,0 +1,73 @@
+/*
+ *
+ *  Copyright 2017 Netflix, Inc.
+ *
```
zhljen: 2018

metacat-common-server/src/main/java/com/netflix/metacat/common/server/properties/Definition.java
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Definition metadata related properties.
+ *
+ * @author amajumdar
+ * @since 1.1.1
+ */
+@lombok.Data
+public class Definition {
+
```
zhljen: This class is named as 'Definition'  is strange, which doesn't tell much about itself.

zhljen: Maybe DefinitionMetadataProperties ?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/properties/PropertyUtils.java
```diff
@@ -62,4 +64,26 @@ protected PropertyUtils() {
...
+
+    /**
+     * Convert a delimited string into a Set of {@code QualifiedName}.
+     *
+     * @param names     The list of names to split
+     * @param delimiter The delimiter to use for splitting
+     * @return The list of qualified names
+     */
+    static Set<QualifiedName> delimitedStringsToQualifiedNamesSet(
+        @Nonnull @NonNull final String names,
```
zhljen: I thought we decided to flip to the other way, i.e. making Nonnull as default

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -184,15 +189,48 @@ public TableDto deleteAndReturn(final QualifiedName name, final boolean isMView)
...
+     * 1. If the system is configured to delete deifnition metadata.
+     * 2. If the system is configured not to but the tableName is configured to either explicitly or if the
+     * table's database/catalog is configure to.
+     * @param tableName table name
+     * @return whether or not to delete definition metadata
+     */
+    private boolean canDeleteMetadata(final QualifiedName tableName) {
+        return config.canDeleteTableDefinitionMetadata() || isEnabledForTableDefinitionMetadataDelete(tableName);
+    }
+
```
zhljen: Can we rename `canDeleteTableDefinitionMetadata()` as `forceTableDefinitionMetadataEnabled() `? looks like this flag forcing all the table defmetadata to be deleted.
and 
`isEnabledForTableDefinitionMetadataDelete(tableName)` as `tableDefinitionMetadataDeleteEnabled(tableName)` this flag is individual table flag

metacat-main/src/test/groovy/com/netflix/metacat/main/services/impl/TableServiceImplSpec.groovy
```diff
@@ -0,0 +1,112 @@
...
+import com.netflix.metacat.testdata.provider.DataDtoProvider
+import com.netflix.spectator.api.NoopRegistry
+import spock.lang.Specification
+
+/**
+ * Tests for the TableServiceImpl.
+ *
+ * @author amajumdar
+ * @since 1.2.0
+ */
```
zhljen: I am not sure if this should be 1.2.0 or 1.1.1. Some codes are marked as 1.1.1 previously, and we should make them consistent as they are from the same change,

metacat-common-server/src/main/java/com/netflix/metacat/common/server/properties/Definition.java
ajoymajumdar: Will change.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/properties/Definition.java
ajoymajumdar: It is under properties package and also we have followed this convention for other properties as well. This class refers to property metacat.definition.metadata...

metacat-common-server/src/main/java/com/netflix/metacat/common/server/properties/PropertyUtils.java
ajoymajumdar: Kept it consistent with the other method in the class. We should do the massive change for all classes.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
ajoymajumdar: I disagree with the rename. It could either be kept as canDeleteTableDefinitionMetadata or changed to isDeleteDefinitionMetadataEnabled. The second configuration is used as a whitelist for any qualified name.

metacat-main/src/test/groovy/com/netflix/metacat/main/services/impl/TableServiceImplSpec.groovy
ajoymajumdar: Yes, all these changes will go as 1.2.0. I will change where ever it says 1.1.1

PR #221, opened 2018-01-23
Adding base.time to pig type converter

zhljen:


zhljen:
Changed based on comments.

PR #228, opened 2018-02-06
Handling null pointer in GetPartitionRequestDto

zhljen:


metacat-main/src/main/java/com/netflix/metacat/main/services/impl/PartitionServiceImpl.java
```diff
@@ -137,12 +137,8 @@ public PartitionServiceImpl(
         final GetPartitionsRequestDto getPartitionsRequestDto
     ) {
 
-        String filterExpression = null;
-        List<String> partitionNames = null;
-        if (getPartitionsRequestDto != null) {
-            filterExpression = getPartitionsRequestDto.getFilter();
-            partitionNames = getPartitionsRequestDto.getPartitionNames();
-        }
+        final String filterExpression = getPartitionsRequestDto.getFilter();
```
ajoymajumdar: Why did we remove the null check?

zhljen: Agreed. will make the change.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/PartitionServiceImpl.java
```diff
@@ -134,29 +135,28 @@ public PartitionServiceImpl(
...
+            converterUtil.toPartitionListRequest(getPartitionsRequestDto, pageable, sort);
 
-        String filterExpression = null;
-        List<String> partitionNames = null;
-        if (getPartitionsRequestDto != null) {
-            filterExpression = getPartitionsRequestDto.getFilter();
-            partitionNames = getPartitionsRequestDto.getPartitionNames();
-        }
+        final ConnectorRequestContext connectorRequestContext = converterUtil.toConnectorContext(metacatRequestContext);
+        final List<PartitionInfo> resultInfo = service
```
ajoymajumdar: This should come after the validation check that throws IllegalArugumentException.

zhljen: Sounds right.

PR #233, opened 2018-02-19
Getting the docker-compose example working on Windows MinGW / MSys

brosenfi:
These are some changes I made in order to get the docker-compose example working on windows in a MinGW / MSys bash terminal.

The change to accept Lf / Cr / CrLf line-endings is nice to have when shutting auto-crlf conversion off in windows - which is a general practice when having projects build docker containers (script creation needs to have Lf line-endings when adding to linux-based containers).

Use of the 'sh' command tells MSys explicitly this is a shell-script and it can use the MSys local binary for that.

The other two changes (in world.sql and start_metacat_test_cluster.sh) were required within the container environment to get the example running - not sure why this was any different within Windows than *nix machines.

There is also a few groovy tests that are failing for the Hive connector build (within PartitionUtilSpec.groovy) that I see are not failing on *nix machines - I haven't fixed that yet though (just skipping tests on the build at this point).

### DIFF
--- a/codequality/checkstyle/checkstyle.xml
+++ b/codequality/checkstyle/checkstyle.xml
@@ -18,7 +18,9 @@
     </module>
 
     <!-- Checks that there is a newline at the end of each file. -->
-    <module name="NewlineAtEndOfFile"/>
+    <module name="NewlineAtEndOfFile">
+		<property name="lineSeparator" value="lf_cr_crlf"/>
+	</module>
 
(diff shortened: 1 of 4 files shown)


brosenfi:
Build likely failed due to that cassandra init.cql reference - apologies, I didn't see the local volume reference in the docker-compose yaml - I will revert that part specifically and figure out why it wasn't mounting on my test.

PR #235, opened 2018-02-21
Renaming apis from using metadatas to metadata

zhljen:


metacat-common-server/src/main/java/com/netflix/metacat/common/server/usermetadata/UserMetadataService.java
```diff
@@ -52,7 +52,7 @@
      *
      * @param uris list of uris.
      */
-    default void deleteDataMetadatas(List<String> uris) {
+    default void deleteDataMetadata(List<String> uris) {
```
ajoymajumdar: Why this change from plural to singular? This method is deleting mutiple data metadatas.

zhljen: the plural of metadata is metadata..

ajoymajumdar: Do we really need this change?

zhljen: I separated out as a new pull from the previous one. We should fix this name.

PR #237, opened 2018-02-24
Added support to have one catalog name pointing to multiple data store shards.

ajoymajumdar:
Example: We can add multiple catalogs, referencing multiple datasources, addressed by the same catalog name.

### DIFF
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatDeleteTablePartitionPostEvent.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatDeleteTablePartitionPostEvent.java
@@ -19,6 +19,7 @@ package com.netflix.metacat.common.server.events;
 
 import com.netflix.metacat.common.MetacatRequestContext;
 import com.netflix.metacat.common.QualifiedName;
+import com.netflix.metacat.common.dto.PartitionDto;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.NonNull;
@@ -27,6 +28,7 @@ import lombok.ToString;
 import javax.annotation.Nonnull;
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
 
 /**
  * Post table partition delete event.
@@ -35,8 +37,8 @@ import java.util.List;
 @EqualsAndHashCode(callSuper = true)
 @ToString(callSuper = true)
 public class MetacatDeleteTablePartitionPostEvent extends MetacatEvent {
-
     private final List<String> partitionIds;
+    private final List<PartitionDto> partitions;
 
     /**
      * Constructor.
@@ -44,15 +46,17 @@ public class MetacatDeleteTablePartitionPostEvent extends MetacatEvent {
      * @param name           name
      * @param requestContext context
      * @param source         The source object which threw this event
-     * @param partitionIds   partition names
+     * @param partitions  partition dtos
      */
     public MetacatDeleteTablePartitionPostEvent(
         @Nonnull @NonNull final QualifiedName name,
         @Nonnull @NonNull final MetacatRequestContext requestContext,
         @Nonnull @NonNull final Object source,
-        @Nonnull @NonNull final List<String> partitionIds
+        @Nonnull @NonNull final List<PartitionDto> partitions
     ) {
         super(name, requestContext, source);
-        this.partitionIds = Collections.unmodifiableList(partitionIds);
+        this.partitions = Collections.unmodifiableList(partitions);
+        this.partitionIds = partitions.stream()
+            .map(dto -> dto.getName().getPartitionName()).collect(Collectors.toList());
     }
 }
--- a/metacat-main/src/main/java/com/netflix/metacat/main/services/MetacatServiceHelper.java
+++ b/metacat-main/src/main/java/com/netflix/metacat/main/services/MetacatServiceHelper.java
@@ -239,7 +239,7 @@ public class MetacatServiceHelper {
                     name,
                     metacatRequestContext,
                     this,
-                    Lists.newArrayList(name.getPartitionName())
+                    Lists.newArrayList(PartitionDto.builder().name(name).build())
                 )
             );
         } else if (name.isViewDefinition()) {
--- a/metacat-main/src/main/java/com/netflix/metacat/main/services/impl/PartitionServiceImpl.java
+++ b/metacat-main/src/main/java/com/netflix/metacat/main/services/impl/PartitionServiceImpl.java
@@ -57,6 +57,7 @@ import com.netflix.spectator.api.Registry;
 import lombok.extern.slf4j.Slf4j;
 
 import javax.annotation.Nullable;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
@@ -282,6 +283,7 @@ public class PartitionServiceImpl implements PartitionService {
         final QualifiedName name, final List<PartitionDto> partitionDtos) {
         final ConnectorRequestContext connectorRequestContext = converterUtil.toConnectorContext(metacatRequestContext);
         List<HasMetadata> deletePartitions = Lists.newArrayList();
+        List<PartitionDto> deletePartitionDtos = Lists.newArrayList();
         registry.distributionSummary(
             this.partitionAddDistSummary.withTags(name.parts())).record(partitionDtos.size());
         final List<String> partitionIdsForDeletes = dto.getPartitionIdsForDeletes();
@@ -294,8 +296,9 @@ public class PartitionServiceImpl implements PartitionService {
             final List<PartitionInfo> deletePartitionInfos = service.getPartitions(connectorRequestContext, name,
                 converterUtil.toPartitionListRequest(requestDto, null, null));
             if (deletePartitionInfos != null) {
-                deletePartitions = deletePartitionInfos.stream().map(converterUtil::toPartitionDto)
-                    .collect(Collectors.toList());
+                deletePartitionDtos = deletePartitionInfos.stream()
+                    .map(converterUtil::toPartitionDto).collect(Collectors.toList());
+                deletePartitions = new ArrayList<>(deletePartitions);
             }
         }
 
@@ -322,7 +325,7 @@ public class PartitionServiceImpl implements PartitionService {
         if (partitionIdsForDeletes != null && !partitionIdsForDeletes.isEmpty()) {
             eventBus.postAsync(
                 new MetacatDeleteTablePartitionPostEvent(name,
-                    metacatRequestContext, this, partitionIdsForDeletes));
+                    metacatRequestContext, this, deletePartitionDtos));
         }
         eventBus.postAsync(
             new MetacatSaveTablePartitionPostEvent(name, metacatRequestContext, this, partitionDtos, result));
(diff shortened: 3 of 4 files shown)


metacat-common-server/src/main/java/com/netflix/metacat/common/server/spi/MetacatCatalogConfig.java
```diff
@@ -104,6 +89,17 @@ public boolean isThriftInterfaceRequested() {
...
     public static class Keys {
+        /**
+         * Catalog name. Usually catalog name is derived from the name of the catalog properties file.
+         * One could also specify it in the properties under the property name <code>catalog.name</code>.
+         * For example, if a catalog has two shards with catalog defined in two property files,
+         * then we can have the <code>catalog.name</code> set to one name.
+         */
+        public static final String CATALOG_NAME = "catalog.name";
+        /**
+         * Connector type.
```
zhljen: is this type or name?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorContext.java
```diff
@@ -30,6 +30,14 @@
...
      */
     private final String catalogName;
+    /**
+     * Catalog shard name.
+     */
+    private final String catalogShardName;
+    /**
+     * Catalog type.
+     */
+    private final String type;
```
zhljen: We should name the type explicitly as connectorType.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/util/DataSourceManager.java
```diff
@@ -50,13 +50,13 @@ public static DataSourceManager get() {
...
-     * @param catalogName catalog name
+     * @param name catalog name
      * @param properties  properties
      * @return DataSourceManager
      */
-    public DataSourceManager load(final String catalogName, final Map<String, String> properties) {
-        if (dataSources.get(catalogName) == null) {
-            createDataSource(catalogName, properties);
+    public DataSourceManager load(final String name, final Map<String, String> properties) {
+        if (dataSources.get(name) == null) {
```
zhljen: what is this name? If this is not 'catalogName' anymore, we should change the comments 
`* @param name catalog name` to the right one.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/util/DataSourceManager.java
```diff
@@ -78,15 +78,15 @@ public DataSourceManager load(final String catalogName, final Properties propert
     /**
      * Returns the data source loaded for the given catalog.
      *
-     * @param catalogName catalog name
+     * @param name catalog name
      * @return DataSource
```
zhljen: same as above.

metacat-main/src/main/java/com/netflix/metacat/main/manager/CatalogManager.java
```diff
@@ -38,7 +38,15 @@
...
 
 /**
- * Catalog manager.
+ * Catalog manager. This loads the catalogs defined as .properties files under the location defined by config property
+ * <code>metacat.plugin.config.location</code>.
+ * Usually there is a one-to-one mapping between a catalog and a data store. We could also have data stores
+ * (mostly sharded) addressed by a single catalog name.
+ * If a data store is sharded and needs to be represented by a single catalog name, then we will have multiple catalog
+ * property files, each referencing to its physical data store and having the same <code>catalog.name</code>.
+ * In this case, <code>catalog.name</code> and <code>metacat.schema.whitelist</code> will be used to point to the right
```
zhljen: schema. whitelist?  Why not using 'shards'?

metacat-main/src/main/java/com/netflix/metacat/main/manager/CatalogManager.java
```diff
@@ -114,16 +124,22 @@ private void loadCatalog(final File file) throws Exception {
...
+        // If multiple catalog property files refer the same catalog name, then there will be multiple shard names
+        // with the same catalog name.
+        final String catalogShardName = Files.getNameWithoutExtension(file.getName());
+        // If catalog name is not specified, then use the catalog shard name.
+        final String catalogName = properties.getOrDefault(MetacatCatalogConfig.Keys.CATALOG_NAME, catalogShardName);
+        final ConnectorContext connectorContext =
+            new ConnectorContext(catalogName, catalogShardName, connectorType, config, registry, properties);
+        this.connectorManager.createConnection(connectorContext);
         log.info("-- Added catalog {} using connector {} --", catalogName, connectorType);
     }
```
zhljen: are we allowing one file having multiple shards configs? Also, can we log the shard explicitly? 
`log.info("-- Added catalog {} using connector {} --", catalogName, connectorType);`

metacat-main/src/main/java/com/netflix/metacat/main/manager/ConnectorManager.java
```diff
@@ -103,31 +116,91 @@ public void addPlugin(final ConnectorPlugin connectorPlugin) {
...
-
             final MetacatCatalogConfig catalogConfig =
-                MetacatCatalogConfig.createFromMapAndRemoveProperties(connectorType,
+                MetacatCatalogConfig.createFromMapAndRemoveProperties(connectorType, catalogName,
                     connectorContext.getConfiguration());
-            catalogs.put(catalogName, catalogConfig);
+            final List<String> databaseNames = catalogConfig.getSchemaWhitelist();
+            if (databaseNames.isEmpty()) {
+                Preconditions.checkState(!catalogs.contains(catalogName, EMPTY_STRING),
+                    "A catalog with name %s already exists", catalogName);
```
zhljen: What if the catalog doesn't exist? Should we logging error for this case?

metacat-main/src/main/java/com/netflix/metacat/main/manager/ConnectorManager.java
```diff
@@ -103,31 +116,91 @@ public void addPlugin(final ConnectorPlugin connectorPlugin) {
...
+                log.debug("Catalog {} doesn't support getTableService. Ignoring.", catalogName);
+            }
+            try {
+                partitionServices.add(connectorFactory.getPartitionService());
+            } catch (UnsupportedOperationException e) {
+                log.debug("Catalog {} doesn't support getPartitionService. Ignoring.", catalogName);
+            }
+            final CatalogHolder catalogHolder = new CatalogHolder(catalogConfig, connectorFactory);
+            if (databaseNames.isEmpty()) {
+                catalogs.put(catalogName, EMPTY_STRING, catalogHolder);
```
zhljen: Maybe this should be moved to the top along with the first
`if (databaseNames.isEmpty()) {`

metacat-main/src/main/java/com/netflix/metacat/main/manager/ConnectorManager.java
```diff
@@ -41,26 +42,39 @@
...
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
 
 /**
  * Connector manager.
  */
 @Slf4j
 public class ConnectorManager {
+    private static final String EMPTY_STRING = "";
```
zhljen: Actually. this (ConnectorManager) looks more like a catalog manager and the catalogManger is more like a connectorManager. We should switch their name.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorFactory.java
```diff
@@ -57,12 +57,19 @@ default ConnectorPartitionService getPartitionService() {
     }
 
     /**
-     * Returns the name of the connector.
+     * Returns the name of the catalog.
      *
-     * @return Returns the name of the connector.
+     * @return Returns the name of the catalog.
      */
     String getName();
```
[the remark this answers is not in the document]
ajoymajumdar: Consider Connector as the definition and Catalog has an instance of it. Multiple catalogs can have one connector. I have kept the method name as it is but modified the comment to clear the confusion that was raised in our discussion.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorContext.java
ajoymajumdar: Will do.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/spi/MetacatCatalogConfig.java
ajoymajumdar: This is the connector type. We had this property as connector.name. Probably we had this inherited from presto.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/util/DataSourceManager.java
ajoymajumdar: Can be anything, this is a generic class. Consumers can use any name - catalog name or catalog shard name.

metacat-main/src/main/java/com/netflix/metacat/main/manager/CatalogManager.java
ajoymajumdar: This was already defined. Concept of shards is higher than this. A shard will contain a list of databases.

metacat-main/src/main/java/com/netflix/metacat/main/manager/ConnectorManager.java
ajoymajumdar: CatalogManager loads catalogs. ConnectorManager provides the right connector properties/services.

metacat-main/src/main/java/com/netflix/metacat/main/manager/ConnectorManager.java
zhljen: this is empty represents everything ( as default) or a real empty database?  Are we use empty string as a placeholder for the catalogs having no database?

metacat-main/src/main/java/com/netflix/metacat/main/manager/ConnectorManager.java
```diff
@@ -238,4 +328,14 @@ public ConnectorTypeConverter getTypeConverter(final String connectorType) {
     public ConnectorInfoConverter getInfoConverter(final String connectorType) {
         return getPlugin(connectorType).getInfoConverter();
     }
+
+    /**
+     * A Holder class holding the catalog's config and connector factory.
+     */
+    @Data
+    @AllArgsConstructor
+    private static class CatalogHolder {
```
zhljen: I feel we should factor out the catalogRelated operation into a new class, separating from the database/table/part services related informations. The catalog class has 
`private final Table<String, String, CatalogHolder> catalogs = HashBasedTable.create();`
and catalog related information.

zhljen: The catalogHolder is not needed if the table has four columns: catalogName, databaseName, config, connectorFactory

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/CatalogServiceImpl.java
```diff
@@ -70,20 +71,29 @@ public CatalogServiceImpl(
...
-                .collect(Collectors.toList())
-        );
+        final List<String> databases = Lists.newArrayList();
+        configs.forEach(config -> {
+            QualifiedName qName = name;
+            if (config.getSchemaWhitelist().isEmpty()) {
+                result.setType(config.getType());
+            } else {
+                qName = QualifiedName.ofDatabase(name.getCatalogName(), config.getSchemaWhitelist().get(0));
+            }
```
zhljen: Don't understand here. If always get the first one, why do we need list?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorFactory.java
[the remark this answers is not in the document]
zhljen: I fee Connector is a mechanism to load catalog, which is more implementation details. Catalog is the concept/definition existing in the metacat architecture.

PR #238, opened 2018-02-27
Changing the partition deletion events to have partitionDtos.

zhljen:


metacat-main/src/main/java/com/netflix/metacat/main/services/MetacatServiceHelper.java
```diff
@@ -108,7 +108,7 @@ public MetacatService getService(final QualifiedName name) {
      * @param metacatRequestContext context
      * @param dto                   dto
      */
-    public void postPreUpdateEvent(
+    public void postPreMetadataUpdateEvent(
```
ajoymajumdar: Renaming is not right.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/PartitionServiceImpl.java
```diff
@@ -365,7 +370,7 @@ public void delete(final QualifiedName name, final List<String> partitionIds) {
                 deleteMetadatas(metacatRequestContext.getUserName(), partitions);
             }
             eventBus.postAsync(
-                new MetacatDeleteTablePartitionPostEvent(name, metacatRequestContext, this, partitionIds)
+                new MetacatDeleteTablePartitionPostEvent(name, metacatRequestContext, this, partitionDtos)
```
ajoymajumdar: On delete we can just publish the partition ids. DO we need to publish the whole partition objects?

zhljen:
Changes to address the review comments.

PR #240, opened 2018-02-27
Added support to have one catalog name pointing to multiple data store shards.

ajoymajumdar:
Example: We can add multiple catalogs, referencing multiple datasources, addressed by the same catalog name.

### DIFF
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorFactory.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorFactory.java
@@ -61,7 +61,7 @@ public interface ConnectorFactory {
      *
      * @return Returns the name of the catalog.
      */
-    String getName();
+    String getCatalogName();
 
     /**
      * Returns the name of the catalog shard.
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorFactory.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorFactory.java
@@ -57,11 +57,18 @@ public interface ConnectorFactory {
     }
 
     /**
-     * Returns the name of the connector.
+     * Returns the name of the catalog.
      *
-     * @return Returns the name of the connector.
+     * @return Returns the name of the catalog.
      */
     String getName();
+
+    /**
+     * Returns the name of the catalog shard.
+     *
+     * @return Returns the name of the catalog shard.
+     */
+    String getCatalogShardName();
 
     /**
      * Shuts down the factory.
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/SpringConnectorFactory.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/SpringConnectorFactory.java
@@ -30,20 +30,24 @@ import org.springframework.core.env.StandardEnvironment;
  */
 
 public abstract class SpringConnectorFactory implements ConnectorFactory {
-    protected final String catalogName;
     protected final AnnotationConfigApplicationContext ctx;
+    private final String catalogName;
+    private final String catalogShardName;
 
     /**
      * Constructor.
      *
      * @param catalogName            catalog name
+     * @param catalogShardName       catalog shard name
      * @param connectorInfoConverter connector info converter
      * @param connectorContext       connector related config
      */
     public SpringConnectorFactory(final String catalogName,
+                                  final String catalogShardName,
                                   final ConnectorInfoConverter connectorInfoConverter,
                                   final ConnectorContext connectorContext) {
         this.catalogName = catalogName;
+        this.catalogShardName = catalogShardName;
         this.ctx = new AnnotationConfigApplicationContext();
         this.ctx.setEnvironment(new StandardEnvironment());
         this.ctx.getBeanFactory().registerSingleton("ConnectorContext", connectorContext);
@@ -89,5 +93,13 @@ public abstract class SpringConnectorFactory implements ConnectorFactory {
     @Override
     public String getName() {
         return this.catalogName;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String getCatalogShardName() {
+        return catalogShardName;
     }
 }
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/SpringConnectorFactory.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/SpringConnectorFactory.java
@@ -91,7 +91,7 @@ public abstract class SpringConnectorFactory implements ConnectorFactory {
      * {@inheritDoc}
      */
     @Override
-    public String getName() {
+    public String getCatalogName() {
         return this.catalogName;
     }
 
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/SpringConnectorFactory.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/SpringConnectorFactory.java
@@ -37,17 +37,13 @@ public abstract class SpringConnectorFactory implements ConnectorFactory {
     /**
      * Constructor.
      *
-     * @param catalogName            catalog name
-     * @param catalogShardName       catalog shard name
      * @param connectorInfoConverter connector info converter
      * @param connectorContext       connector related config
      */
-    public SpringConnectorFactory(final String catalogName,
-                                  final String catalogShardName,
-                                  final ConnectorInfoConverter connectorInfoConverter,
+    public SpringConnectorFactory(final ConnectorInfoConverter connectorInfoConverter,
                                   final ConnectorContext connectorContext) {
-        this.catalogName = catalogName;
-        this.catalogShardName = catalogShardName;
+        this.catalogName = connectorContext.getCatalogName();
+        this.catalogShardName = connectorContext.getCatalogShardName();
         this.ctx = new AnnotationConfigApplicationContext();
         this.ctx.setEnvironment(new StandardEnvironment());
         this.ctx.getBeanFactory().registerSingleton("ConnectorContext", connectorContext);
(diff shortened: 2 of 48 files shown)


metacat-main/src/main/java/com/netflix/metacat/main/manager/ConnectorManager.java
```diff
@@ -41,26 +42,39 @@
...
+     * catalogs addressed by the same <code>catalog.name</code> pointing to shards of a data store, then there will be
+     * multiple entries in this table. An entry will be identified using the catalog name and the database name.
+     */
+    private final Table<String, String, CatalogHolder> catalogs = HashBasedTable.create();
+    private final Set<MetacatCatalogConfig> catalogConfigs = Sets.newHashSet();
+    private final Set<ConnectorDatabaseService> databaseServices = Sets.newHashSet();
+    private final Set<ConnectorTableService> tableServices = Sets.newHashSet();
+    private final Set<ConnectorPartitionService> partitionServices = Sets.newHashSet();
     private final AtomicBoolean stopped = new AtomicBoolean();
     private final Config config;
```
zhljen: I feel we should factor out the catalogRelated operation into a new class, separating from the database/table/part services related informations. The catalog class has 
`private final Table<String, String, CatalogHolder> catalogs = HashBasedTable.create();`
and catalog related information.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorFactory.java
```diff
@@ -57,11 +57,18 @@ default ConnectorPartitionService getPartitionService() {
...
      */
-    String getName();
+    String getCatalogName();
+
+    /**
+     * Returns the name of the catalog shard.
+     *
+     * @return Returns the name of the catalog shard.
+     */
+    String getCatalogShardName();
```
tgianos: Should this return an `Optional` to reflect that it could be null or empty?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/SpringConnectorFactory.java
```diff
@@ -30,20 +30,24 @@
...
     /**
      * Constructor.
      *
      * @param catalogName            catalog name
+     * @param catalogShardName       catalog shard name
      * @param connectorInfoConverter connector info converter
      * @param connectorContext       connector related config
      */
     public SpringConnectorFactory(final String catalogName,
+                                  final String catalogShardName,
```
tgianos: If this can be null I'd recommend flagging it with `@Nullable`

metacat-common-server/src/main/java/com/netflix/metacat/common/server/spi/MetacatCatalogConfig.java
```diff
@@ -49,10 +53,11 @@ private MetacatCatalogConfig(
      * Creates the config.
      *
      * @param type       type
+     * @param catalogName catalog name
      * @param properties properties
      * @return config
      */
-    public static MetacatCatalogConfig createFromMapAndRemoveProperties(final String type,
+    public static MetacatCatalogConfig createFromMapAndRemoveProperties(final String type, final String catalogName,
```
tgianos: You should probably either put all of these on new lines or keep them all in one line. Otherwise it's easy to miss seeing parameters.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/spi/MetacatCatalogConfig.java
```diff
@@ -104,6 +89,17 @@ public boolean isThriftInterfaceRequested() {
...
     public static class Keys {
+        /**
+         * Catalog name. Usually catalog name is derived from the name of the catalog properties file.
+         * One could also specify it in the properties under the property name <code>catalog.name</code>.
+         * For example, if a catalog has two shards with catalog defined in two property files,
+         * then we can have the <code>catalog.name</code> set to one name.
+         */
+        public static final String CATALOG_NAME = "catalog.name";
+        /**
+         * Connector type.
```
tgianos: This comment doesn't seem to match the name of the variable?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/util/DataSourceManager.java
```diff
@@ -50,13 +50,13 @@ public static DataSourceManager get() {
...
-     * @param catalogName catalog name
+     * @param name catalog name
      * @param properties  properties
      * @return DataSourceManager
      */
-    public DataSourceManager load(final String catalogName, final Map<String, String> properties) {
-        if (dataSources.get(catalogName) == null) {
-            createDataSource(catalogName, properties);
+    public DataSourceManager load(final String name, final Map<String, String> properties) {
+        if (dataSources.get(name) == null) {
```
tgianos: Should this be synchronized instead of synchronizing the method below? that way you avoid having two threads waiting on that method just to do the check again?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/SpringConnectorFactory.java
```diff
@@ -30,20 +30,24 @@
  */
 
 public abstract class SpringConnectorFactory implements ConnectorFactory {
-    protected final String catalogName;
     protected final AnnotationConfigApplicationContext ctx;
+    private final String catalogName;
```
tgianos: It seems to me rather than passing `catalogName` and `catalogShardName` to every class through the whole system (with the potential to want to pass more information later) it would be better to abstract these behind a class which encapsulates them so you only have to pass one object around and it can easily be extended without touching as many classes next time

metacat-main/src/main/java/com/netflix/metacat/main/manager/CatalogManager.java
```diff
@@ -38,7 +38,15 @@
...
 /**
- * Catalog manager.
+ * Catalog manager. This loads the catalogs defined as .properties files under the location defined by config property
+ * <code>metacat.plugin.config.location</code>.
+ * Usually there is a one-to-one mapping between a catalog and a data store. We could also have data stores
+ * (mostly sharded) addressed by a single catalog name.
+ * If a data store is sharded and needs to be represented by a single catalog name, then we will have multiple catalog
+ * property files, each referencing to its physical data store and having the same <code>catalog.name</code>.
+ * In this case, <code>catalog.name</code> and <code>metacat.schema.whitelist</code> will be used to point to the right
+ * data store. A catalog with no <code>metacat.schema.whitelist</code> setting will be the default catalog representing
```
tgianos: Wouldn't it just be cleaner to have a property representing which is the default? `catalog.default=true`?
Also why isn't the shardname represented in this description?

metacat-main/src/main/java/com/netflix/metacat/main/manager/CatalogManager.java
```diff
@@ -114,17 +124,23 @@ private void loadCatalog(final File file) throws Exception {
...
             connectorType != null,
             "Catalog configuration %s does not contain connector.name",
             file.getAbsoluteFile()
         );
 
-        final String catalogName = Files.getNameWithoutExtension(file.getName());
-        final ConnectorContext connectorContext = new ConnectorContext(catalogName, config, registry, properties);
-        this.connectorManager.createConnection(catalogName, connectorType, connectorContext);
-        log.info("-- Added catalog {} using connector {} --", catalogName, connectorType);
+        // Catalog shard name should be unique. Usually the catalog name is same as the catalog shard name.
```
tgianos: This seems overly complex why don't we just have 3 properties:
`catalog.name`,
`catalog.shardName` (optional)
`catalog.default` (default false)

metacat-main/src/main/java/com/netflix/metacat/main/manager/ConnectorManager.java
```diff
@@ -103,31 +116,91 @@ public void addPlugin(final ConnectorPlugin connectorPlugin) {
...
+                    Preconditions.checkState(!catalogs.contains(catalogName, databaseName),
+                        "A catalog with name %s for database %s already exists", catalogName, databaseName);
+                });
+            }
+            catalogConfigs.add(catalogConfig);
+            final ConnectorFactory connectorFactory = connectorPlugin.create(connectorContext);
+            try {
+                databaseServices.add(connectorFactory.getDatabaseService());
+            } catch (UnsupportedOperationException e) {
+                log.debug("Catalog {} doesn't support getDatabaseService. Ignoring.", catalogName);
```
tgianos: These error messages don't reflect possibility of a shard name?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorFactory.java
ajoymajumdar: This will be the same as catalogName when there is only one shard for one catalog.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/util/DataSourceManager.java
ajoymajumdar: No, you always want to synchronize at a level closest to the update of the critical data. Putting synchronized in this method also would make the threads wait to get the lock.

metacat-main/src/main/java/com/netflix/metacat/main/manager/CatalogManager.java
ajoymajumdar: It is a design choice whether to use one property or multiple properties.

metacat-main/src/main/java/com/netflix/metacat/main/manager/ConnectorManager.java
ajoymajumdar: Yeah, we could pass in the shard name but since we are ignoring it, I did not add it.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/SpringConnectorFactory.java
ajoymajumdar: No, it is never null.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/spi/MetacatCatalogConfig.java
ajoymajumdar: It is the connector type.

metacat-main/src/main/java/com/netflix/metacat/main/manager/ConnectorManager.java
ajoymajumdar: Will add.

PR #241, opened 2018-03-06
Changes to add interceptor in usermetadata services

zhljen:


metacat-main/src/main/java/com/netflix/metacat/main/services/GetDatabaseParameters.java
```diff
@@ -0,0 +1,34 @@
...
+
+import lombok.Builder;
+import lombok.Data;
+
+/**
+ * Getdatabase Parameters.
+ * @author zhenl
+ * @since 1.2.0
+ */
+@Data
```
tgianos: Take a look at Lombok's [Value](https://projectlombok.org/features/Value)

metacat-main/src/main/java/com/netflix/metacat/main/api/v1/MetacatController.java
```diff
@@ -578,7 +580,10 @@ public DatabaseDto getDatabase(
         return this.requestWrapper.processRequest(
             name,
             "getDatabase",
-            () -> databaseService.get(name, includeUserMetadata, includeTableNames)
+            () -> databaseService.get(name,
+                GetDatabaseParameters.builder()
```
ajoymajumdar: Isn't GetDatabaseRequest more appropriate?

metacat-main/src/main/java/com/netflix/metacat/main/api/v1/MetacatController.java
```diff
@@ -1035,7 +1047,7 @@ public TableDto updateMView(
         }
     )
     @Override
-    public TableDto updateTable(
+    public TableDto  updateTable(
```
ajoymajumdar: Why is there a double-spacing? Do you have the right code formatter?

metacat-main/src/main/java/com/netflix/metacat/main/api/v1/MetacatController.java
zhljen: This is the internal parameter. We have the getTableRequest xx like class for the calls from outside.

metacat-main/src/main/java/com/netflix/metacat/main/services/GetDatabaseParameters.java
zhljen: Will change to `@Value`, which makes senses here.

metacat-main/src/main/java/com/netflix/metacat/main/api/v1/MetacatController.java
zhljen: Should be single-space. I will change it.

metacat-main/src/main/java/com/netflix/metacat/main/api/v1/MetacatController.java
ajoymajumdar: GetDatabaseServiceRequest or GetDatabaseServiceParameters?

zhljen: Sure, that's reasonable. I will rename to GetDatabaseServiceParameters

metacat-main/src/main/java/com/netflix/metacat/main/api/v1/TagController.java
```diff
@@ -309,11 +320,20 @@ public void removeTableTags(
                     throw new TableNotFoundException(name);
                 }
                 final TableDto oldTable = this.tableService
-                    .get(name, true)
+                    .get(name, GetTableParameters.builder()
+                        .includeInfo(true)
+                        .includeDataMetadata(true)
+                        .includeDefinitionMetadata(true)
+                        .disableOnReadMetadataIntercetor(true)
```
ajoymajumdar: The code to use this flag can get complicated. Why didn't we add the interceptor logic at the service layer in metacat-main?

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MysqlUserMetadataInterceptorService.java
```diff
@@ -0,0 +1,167 @@
...
+        super(jdbcTemplate, metacatJson, config);
+        this.metadataInterceptor = metadataInterceptor;
+    }
+
+    @Nonnull
+    @Override
+    @Transactional(readOnly = true)
+    public Optional<ObjectNode> getDefinitionMetadata(
+        @Nonnull final QualifiedName name) {
+        final Optional<ObjectNode> retData = getJsonForKey(
```
ajoymajumdar: why are we not delegating the call to the method in the super class?

metacat-main/src/main/java/com/netflix/metacat/main/api/v1/TagController.java
zhljen: This is more of clean-up of the existing API. We should make the get() API call change regardless, since adding flags doesn't scale and the flags can be set in a wrapper which is hidden. 
I am not sure if you understand what you mean by adding the interceptor logic at service layer. We want to put all the interceptor thing in the userMetadataService.

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MysqlUserMetadataInterceptorService.java
zhljen: Yes, we can. I will change it.

zhljen:
Made the changes to provide unique view ( applying interceptors ) for outside API results.

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchRefresh.java
```diff
@@ -389,7 +391,12 @@ private void deleteUnmarkedEntities(final List<QualifiedName> qNames,
                     boolean result = false;
                     try {
                         unmarkedDatabaseNames.add(databaseDto.getName().toString());
-                        final DatabaseDto dto = databaseService.get(databaseDto.getName(), false, false);
+                        final DatabaseDto dto = databaseService.get(databaseDto.getName(),
+                            GetDatabaseServiceParameters.builder()
+                                .includeUserMetadata(false)
+                                .includeTableNames(false)
+                                .disableOnReadMetadataIntercetor(false)
```
ajoymajumdar: we can disable the interceptor.

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchRefresh.java
```diff
@@ -432,7 +439,13 @@ private void deleteUnmarkedEntities(final List<QualifiedName> qNames,
                     boolean result = false;
                     try {
                         unmarkedTableNames.add(tableDto.getName().toString());
-                        final Optional<TableDto> dto = tableService.get(tableDto.getName(), false);
+                        final Optional<TableDto> dto = tableService.get(tableDto.getName(),
+                            GetTableServiceParameters.builder()
+                                .includeDataMetadata(false)
+                                .disableOnReadMetadataIntercetor(false)
```
ajoymajumdar: we can disable the interceptor.

zhljen:
Adding documents to address comments.

zhljen:
Update and resubmit.

PR #246, opened 2018-03-16
Fixed issue where the getTable API was incorrectly returning TableNotFoundException instead of ConnectorException.

ajoymajumdar:


metacat-connector-mysql/src/main/java/com/netflix/metacat/connector/mysql/MySqlConnectorTableService.java
```diff
@@ -59,6 +60,8 @@ protected void setTableInfoDetails(final Connection connection, final TableInfo
                     tableInfo.setAudit(auditInfo);
                 }
             }
+        } catch (final Exception ignored) {
+            log.debug("Ignoring. Error getting the create_time for table {}", tableName);
```
zhljen: We can log this out explicitly or send to atlas, since some one should take a look why they are missing. If the table should have, they should be fixed.

zhljen: Also, what if the update_time is missing? We should pass in the missing field or make the message more generic to avoid confusion.

ajoymajumdar: For now, I will log it as info and just call it out as audit info instead of create_time.

PR #247, opened 2018-03-23
Adding partitionIds to save delete partition SNS events.

zhljen:


metacat-common/src/main/java/com/netflix/metacat/common/dto/notifications/sns/payloads/TablePartitionsUpdatePayload.java
```diff
@@ -35,19 +37,30 @@
 public class TablePartitionsUpdatePayload {
     private final int numCreatedPartitions;
     private final int numDeletedPartitions;
+    private final List<String> createdPartitionIds;
+    private final List<String> deletedPartitionIds;
+    private final String message;
```
ajoymajumdar: Make this a boolean field representing if the keys have been added or not.

zhljen: I want to give the consumer more information, if it's enabled also attached the number of partitionId. If exceeded, we let them know the threshhold.

metacat-main/src/main/java/com/netflix/metacat/main/services/notifications/sns/SNSNotificationServiceImpl.java
```diff
@@ -109,13 +115,19 @@ public void notifyOfPartitionAddition(final MetacatSaveTablePartitionPostEvent e
         final long timestamp = event.getRequestContext().getTimestamp();
         final String requestId = event.getRequestContext().getId();
         // Publish a global message stating how many partitions were updated for the table to the table topic
+        final TablePartitionsUpdatePayload partitionsUpdatePayload;
+        if (this.config.isSnsNotificationAttachPartitionIdsEnabled() && event.getPartitions() != null) {
```
ajoymajumdar: Move the if logic inside the create method

metacat-main/src/main/java/com/netflix/metacat/main/services/notifications/sns/SNSNotificationServiceImpl.java
```diff
@@ -156,12 +204,20 @@ public void notifyOfPartitionDeletion(final MetacatDeleteTablePartitionPostEvent
         final String name = event.getName().toString();
         final long timestamp = event.getRequestContext().getTimestamp();
         final String requestId = event.getRequestContext().getId();
+        final TablePartitionsUpdatePayload partitionsUpdatePayload;
+        if (this.config.isSnsNotificationAttachPartitionIdsEnabled() && event.getPartitions() != null) {
+            partitionsUpdatePayload = createTablePartitionsUpdatePayload(event.getPartitions(), event);
```
ajoymajumdar: Isn't this wrong? This needs to go in as deleted partitions list?

metacat-main/src/main/java/com/netflix/metacat/main/services/notifications/sns/SNSNotificationServiceImpl.java
zhljen: ya, we can move the config check inside, but not the other one since we separated the events as MetacatDeleteTablePartitionPostEvent and MetacatSaveTablePartitionPostEvent, don't want to check the class then cast inside.

metacat-main/src/main/java/com/netflix/metacat/main/services/notifications/sns/SNSNotificationServiceImpl.java
zhljen: Ouch. this is not right.

PR #248, opened 2018-03-23
Synchronous SNS event publish. 

ajoymajumdar:
If the publish fails, the system falls back to another SNS topic(mostly in another region), if provided.

### DIFF
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/AsyncListener.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/AsyncListener.java
@@ -1,0 +1,32 @@
+/*
+ *  Copyright 2018 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ */
+
+package com.netflix.metacat.common.server.events;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation to mark a listener class whose methods will be processed asynchronously by the event bus.
+ *
+ * @author amajumdar
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface AsyncListener {
+}
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationEventMulticaster.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationEventMulticaster.java
@@ -17,19 +17,23 @@
  */
 package com.netflix.metacat.common.server.events;
 
-import lombok.NonNull;
+import com.google.common.collect.Maps;
+import com.netflix.metacat.common.server.properties.MetacatProperties;
+import com.netflix.metacat.common.server.util.RegistryUtil;
+import com.netflix.spectator.api.Registry;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.context.ApplicationEvent;
 import org.springframework.context.ApplicationListener;
 import org.springframework.context.event.ApplicationEventMulticaster;
 import org.springframework.context.event.SimpleApplicationEventMulticaster;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
 
-import javax.annotation.Nonnull;
+import java.util.Map;
 
 /**
- * Event bus implementation using Springs Event Multicaster. By default, Spring supports sunchronous event publishing.
- * This implementation supports both synchronous and asynchronous event publishing. At present, client explicitly
- * call postAsync or postSync to appropriately publish events.
+ * Event bus implementation using Springs Event Multicaster. By default, Spring supports synchronous event publishing.
+ * This implementation supports both synchronous and asynchronous event publishing. If the listener is annotated with
+ * AsyncListener, the event will be published asynchronously using a separate executor pool.
  * This implementation should be used as the application event multicaster in the Springs context.
  * Synchronous publishing of events is handled by this class and the asynchronous publishing of events is handled by
  * the asyncEventMulticaster.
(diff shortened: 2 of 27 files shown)


metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/AsyncListener.java
```diff
@@ -0,0 +1,16 @@
+package com.netflix.metacat.common.server.events;
```
zhljen: To add license.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationEventMulticaster.java
```diff
@@ -54,27 +54,23 @@ public MetacatApplicationEventMulticaster(
...
-    /**
-     * Post event synchronously.
-     *
-     * @param event event
-     */
-    public void postSync(final ApplicationEvent event) {
+    public void post(final ApplicationEvent event) {
         super.multicastEvent(event);
+        this.asyncEventMulticaster.multicastEvent(event);
     }
```
zhljen: Why multicasts twice here?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationEventMulticaster.java
```diff
@@ -54,27 +54,23 @@ public MetacatApplicationEventMulticaster(
...
-    }
-
-    /**
-     * Post event synchronously.
-     *
-     * @param event event
-     */
-    public void postSync(final ApplicationEvent event) {
+    public void post(final ApplicationEvent event) {
         super.multicastEvent(event);
```
zhljen: This method should named as postAsync if that what it does.

zhljen:
Some initial comments.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/AsyncListener.java
ajoymajumdar: Will do.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationEventMulticaster.java
ajoymajumdar: Earlier the publishers were explicitly calling async or sync methods. Now I have changed to make the listeners register as sync or async.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationEventMulticaster.java
ajoymajumdar: Same comment as above.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationEventMulticaster.java
zhljen: If I understand right. The async multicast will happen after executing the listeners in sync multicast, say if one if the action failed and the async multicast won't happen. I guess this is what you want, is just the code looks a bit weird.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
```diff
@@ -50,29 +50,17 @@ public MetacatEventBus(
...
      *
      * @param event event
      */
-    public void postSync(final ApplicationEvent event) {
-        log.debug("Received request to post an event {} synchronously", event);
-        this.eventSyncCounter.increment();
-        this.applicationEventMulticaster.postSync(event);
+    public void post(final ApplicationEvent event) {
+        log.debug("Received request to post an event {}", event);
+        this.eventPublishCounter.increment();
```
zhljen: you combine these into one counter, what if the second async publish step failed. I guess assumption is that it won't fail.

metacat-main/src/main/java/com/netflix/metacat/main/services/notifications/sns/SNSNotificationServiceImpl.java
```diff
@@ -323,93 +319,108 @@ public void notifyOfTableRename(final MetacatRenameTablePostEvent event) {
...
-        } catch (final Exception e) {
+            //
+            // Publish the event to original SNS topic. If we receive an error from SNS, we will then try publishing
+            // to the fallback topic.
+            //
+            try {
+                publishNotification(arn, message, counterKey);
+            } catch (final Exception exception) {
+                if (fallbackArn != null) {
+                    log.info("Fallback published message to topic {} because of error {}",
```
zhljen: Are we assuming microbot listening on the fallback SNS as well?

zhljen: Also, how can we order the messages between these two SNS? For example, the events of the same table ended up into two SNS, how the consumer know which order to process them?

zhljen: Sounds good to me.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/AsyncListener.java
```diff
@@ -0,0 +1,33 @@
+/*
+ *
+ *  Copyright 2016 Netflix, Inc.
```
tgianos: You added the license but the year is wrong. You should change your intellij settings to auto add license to all new classes and you can have a placeholder I think it's like `${year}` or similar and it will auto sub it in every time you create a new class.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationEventMulticaster.java
```diff
@@ -54,27 +54,23 @@ public MetacatApplicationEventMulticaster(
...
+        this.asyncEventMulticaster.multicastEvent(event);
     }
 
     @Override
     public void addApplicationListener(final ApplicationListener listener) {
-        super.addApplicationListener(listener);
-        asyncEventMulticaster.addApplicationListener(listener);
+        if (listener.getClass().isAnnotationPresent(AsyncListener.class)) {
```
tgianos: So in this instance you're pulling the annotation present anywhere in the class. But what happens if a single class has multiple listeners both synchronous and asynchronous? Can you use reflection here on the method level rather than the class level?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
```diff
@@ -50,29 +50,17 @@ public MetacatEventBus(
...
+        this.applicationEventMulticaster.post(event);
```
tgianos: Would it be better to put the counter increment after the post and then somehow tag it success and failure (try/catch?) so you can alert on failures?

metacat-main/src/main/java/com/netflix/metacat/main/configs/SNSNotificationsConfig.java
```diff
@@ -52,18 +45,14 @@
...
      */
     //TODO: See what spring-cloud-aws would provide automatically...
     @Bean
-    @ConditionalOnMissingBean(AmazonSNSAsync.class)
-    public AmazonSNSAsync amazonSNS(final Config config, final Registry registry) {
-        final ExecutorService executor = Executors.newFixedThreadPool(config.getSnsClientThreadCount(),
-            new ThreadFactoryBuilder().setNameFormat("metacat-sns-pool-%d").build());
-        RegistryUtil.registerThreadPool(registry, "metacat-sns-pool", (ThreadPoolExecutor) executor);
-        return new AmazonSNSAsyncClient(DefaultAWSCredentialsProviderChain.getInstance(), executor);
+    @ConditionalOnMissingBean
```
tgianos: I think you lost the class qualifier here. It helps cause otherwise I think it defaults to looking for a bean named `amazonSNS` and doesn't necessarily care about the class type.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationEventMulticaster.java
ajoymajumdar: Having it at a method level will be an overkill. We can always define a listener class to be asynchronous. If we need any of the methods to be synchronous then we move those into a new class.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventBus.java
ajoymajumdar: This counter is just to monitor the number of calls to publish events. The success/failure cases are being monitored by the individual listeners.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationEventMulticaster.java
```diff
@@ -39,53 +43,79 @@
...
+            executor.getThreadPoolExecutor());
+        result.setTaskExecutor(executor);
+        return result;
     }
 
     @Override
     public void removeApplicationListener(final ApplicationListener listener) {
         super.removeApplicationListener(listener);
-        asyncEventMulticaster.removeApplicationListener(listener);
+        asyncEventMulticasters.values().forEach(aem -> aem.removeApplicationListener(listener));
```
zhljen: Each listener can only belong to one asyncEventMulticaster. Should we get the key first and then remove from that instead of blindly calling remove it from all?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventListenerFactory.java
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.context.ApplicationListener;
+import org.springframework.context.event.EventListenerFactory;
+
+import java.lang.reflect.Method;
+
+/**
+ * This class overrides the DefaultEventListenerFactory in the Spring container.
+ *
+ * @author amajumdar
+ */
```
zhljen: Add `@since`

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationListenerMethodAdapter.java
```diff
@@ -0,0 +1,47 @@
...
+import java.lang.reflect.Method;
+
+/**
+ * This class has been introduced to get access to the targetClass in ApplicationListenerMethodAdapter.
+ *
+ * @author amajumdar
+ */
+public class MetacatApplicationListenerMethodAdapter extends ApplicationListenerMethodAdapter {
+    private final Class<?> targetClass;
+
```
zhljen: Rename MetacatApplicationListenerMethodAdapter -> MetacatApplicationListenerAdapter ?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationEventMulticaster.java
```diff
@@ -39,53 +43,79 @@
...
+            ? ((MetacatApplicationListenerMethodAdapter) listener).getTargetClass() : listener.getClass();
+    }
+
+    private ApplicationEventMulticaster createApplicationEventMultiCaster(final String name) {
+        final SimpleApplicationEventMulticaster result = new SimpleApplicationEventMulticaster();
+        final ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
+        executor.setCorePoolSize(metacatProperties.getEvent().getBus().getExecutor().getThread().getCount());
+        executor.initialize();
+        RegistryUtil.registerThreadPool(registry, "metacat.event.pool." + name,
+            executor.getThreadPoolExecutor());
```
zhljen: We should use simple name for the metric name. Or change the registerThreadPool() to use the name as tag.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationListenerMethodAdapter.java
ajoymajumdar: Why?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatEventListenerFactory.java
ajoymajumdar: Will do.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationListenerMethodAdapter.java
```diff
@@ -0,0 +1,48 @@
...
+import java.lang.reflect.Method;
+
+/**
+ * This class has been introduced to get access to the targetClass in ApplicationListenerMethodAdapter.
+ *
+ * @author amajumdar
+ * @since 1.2.x
+ */
+public class MetacatApplicationListenerMethodAdapter extends ApplicationListenerMethodAdapter {
```
tgianos: Can you not just call `getTargetBean`?

metacat-main/src/main/java/com/netflix/metacat/main/configs/SNSNotificationsConfig.java
```diff
@@ -52,18 +46,14 @@
     /**
      * If SNS notifications are desired and no existing client has been created elsewhere
      * in the application create a default client here.
-     * @param config       The system configuration abstraction to use
-     * @param registry     registry for spectator
+     *
      * @return The configured SNS client
      */
     //TODO: See what spring-cloud-aws would provide automatically...
```
tgianos: Still think this TODO is worthwhile. In Genie 4 we're using this.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatApplicationListenerMethodAdapter.java
ajoymajumdar: Even getTargetBean is protected. So we have to extend the class anyways.

metacat-main/src/main/java/com/netflix/metacat/main/configs/SNSNotificationsConfig.java
ajoymajumdar: Can u point to the code?

tgianos: There's not much code since it's a lot of auto configs and whatnot but... https://github.com/Netflix/genie/blob/master/genie-agent/src/main/java/com/netflix/genie/agent/configs/AwsConfig.java

https://github.com/Netflix/genie/blob/master/genie-agent/src/main/java/com/netflix/genie/GenieAgentApplication.java

Still needs cleanup and testing. Reading docs probably would be more of a help.

ajoymajumdar: Looking at the documentation, I see that it is an umbrella lib for clients to all(RDS, S3, SNS, SQS). I'll keep the TODO :) and make the change later.

PR #253, opened 2018-05-09
Fixing table listing in snowflakes connector.

zhljen:


metacat-connector-snowflake/src/main/java/com/netflix/metacat/connector/snowflake/SnowflakeConnectorTableService.java
```diff
@@ -158,4 +166,26 @@ protected void setTableInfoDetails(final Connection connection, final TableInfo
...
+    @Override
+    protected ResultSet getTables(
+        @Nonnull @NonNull final Connection connection,
+        @Nonnull @NonNull final QualifiedName name,
+        @Nullable final QualifiedName prefix
+    ) throws SQLException {
+        final String schema = name.getDatabaseName();
+        final DatabaseMetaData metaData = connection.getMetaData();
+        return prefix == null || StringUtils.isEmpty(prefix.getTableName())
+            ? metaData.getTables(DSE, schema, null, TABLE_TYPES)
```
ajoymajumdar: Will it always be DSE? The best approach would be to use the schema name set in the catalog properties.

zhljen: that's what I thought. Do we expose the property in oss?

ajoymajumdar: Add this as a config property as part of the catalog configuration file

PR #258, opened 2018-05-18
Implement few thrift APIs

ajoymajumdar:
Implement few thrift APIs - alter_database, drop_database, add_partitions_pspec, get_part_specs_by_filter, get_partitions_pspec and get_partitions_by_expr

### DIFF
--- a/metacat-functional-tests/src/functionalTest/groovy/com/netflix/metacat/MetacatSmokeThriftSpec.groovy
+++ b/metacat-functional-tests/src/functionalTest/groovy/com/netflix/metacat/MetacatSmokeThriftSpec.groovy
@@ -29,10 +29,16 @@ import com.netflix.metacat.testdata.provider.DataDtoProvider
 import org.apache.hadoop.hive.conf.HiveConf
 import org.apache.hadoop.hive.metastore.api.AlreadyExistsException
 import org.apache.hadoop.hive.metastore.api.Database
+import org.apache.hadoop.hive.metastore.api.NoSuchObjectException
+import org.apache.hadoop.hive.ql.exec.FunctionRegistry
 import org.apache.hadoop.hive.ql.metadata.Hive
 import org.apache.hadoop.hive.ql.metadata.Partition
 import org.apache.hadoop.hive.ql.metadata.Table
+import org.apache.hadoop.hive.ql.plan.ExprNodeColumnDesc
+import org.apache.hadoop.hive.ql.plan.ExprNodeConstantDesc
+import org.apache.hadoop.hive.ql.plan.ExprNodeGenericFuncDesc
 import org.apache.hadoop.hive.ql.session.SessionState
+import org.apache.hadoop.hive.serde2.typeinfo.TypeInfoFactory
 import org.apache.log4j.Level
 import org.apache.log4j.Logger
 import org.slf4j.LoggerFactory
@@ -148,6 +154,57 @@ class MetacatSmokeThriftSpec extends Specification {
         client.createDatabase(new Database(databaseName, 'test_db1', null, null))
         then:
         thrown(AlreadyExistsException)
+        cleanup:
+        client.dropDatabase(databaseName)
+        where:
+        client << clients.values()
+        catalogName << clients.keySet()
+    }
+
+    @Unroll
+    def "Test update database for catalog #catalogName"() {
+        given:
+        def invalidDatabaseName = 'test_db1_invalid' + catalogName
+        def databaseName = 'test_db1_' + catalogName
+        client.createDatabase(new Database(databaseName, 'test_db1', null, null))
+        when:
+        client.alterDatabase(databaseName, new Database(databaseName, 'test_db1', null, null))
+        then:
+        def database = client.getDatabase(databaseName)
+        def databaseUri = database.locationUri
+        database != null && database.name == 'test_db1_' + catalogName
+        when:
+        client.alterDatabase(databaseName, new Database(databaseName, 'test_db1', databaseUri + 1, null))
+        then:
+        def uDatabase = client.getDatabase(databaseName)
+        //Database uri in hive does not get updated.
+        uDatabase != null && uDatabase.name == 'test_db1_' + catalogName && uDatabase.locationUri == databaseUri
+        when:
+        client.alterDatabase(invalidDatabaseName, new Database(invalidDatabaseName, 'test_db1', null, null))
+        then:
+        //Hive metsatore does not throw NoSuchObjectException
+        noExceptionThrown()
+        cleanup:
+        client.dropDatabase(databaseName)
+        where:
+        client << clients.values()
+        catalogName << clients.keySet()
+    }
+
+    @Unroll
+    def "Test drop database for catalog #catalogName"() {
+        given:
+        def invalidDatabaseName = 'test_db1_invalid' + catalogName
+        def databaseName = 'test_db1_' + catalogName
+        client.createDatabase(new Database(databaseName, 'test_db1', null, null))
+        when:
+        client.dropDatabase(databaseName)
+        then:
+        client.getDatabase(databaseName) == null
+        when:
+        client.dropDatabase(invalidDatabaseName)
+        then:
+        thrown(NoSuchObjectException)
         where:
         client << clients.values()
         catalogName << clients.keySet()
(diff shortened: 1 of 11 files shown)


metacat-main/src/main/java/com/netflix/metacat/main/api/v1/MetacatController.java
```diff
@@ -956,6 +957,7 @@ public void updateDatabase(
             () -> {
                 final DatabaseDto newDto = new DatabaseDto();
                 newDto.setName(name);
+                newDto.setMetadata(databaseUpdateRequestDto.getMetadata());
                 newDto.setDefinitionMetadata(databaseUpdateRequestDto.getDefinitionMetadata());
```
zhljen: A functional test is needed to check this updateDatabase() since this metadata update operation is new.

metacat-thrift/src/main/java/com/netflix/metacat/thrift/CatalogThriftHiveMetastore.java
```diff
@@ -255,7 +261,18 @@ public int add_partitions(final List<Partition> newParts) throws TException {
```
zhljen: add 
    /**
     * {@inheritDoc}
     */

metacat-thrift/src/main/java/com/netflix/metacat/thrift/CatalogThriftHiveMetastore.java
```diff
@@ -295,7 +312,15 @@ public AddPartitionsResult add_partitions_req(final AddPartitionsRequest request
      */
     @Override
     public void alter_database(final String dbname, final Database db) throws TException {
-        throw unimplemented("alter_database", new Object[]{dbname, db});
+        requestWrapper("update_database", new Object[]{db}, () -> {
+            if (dbname == null || db == null) {
+                throw new InvalidInputException("Invalid database request");
+            }
+            final DatabaseCreateRequestDto dto = new DatabaseCreateRequestDto();
+            dto.setMetadata(db.getParameters());
```
zhljen: I would prefer to use builder pattern, instead of new an object and set the property.

metacat-thrift/src/main/java/com/netflix/metacat/thrift/CatalogThriftHiveMetastore.java
```diff
@@ -295,7 +312,15 @@ public AddPartitionsResult add_partitions_req(final AddPartitionsRequest request
...
+            v1.updateDatabase(catalogName, normalizeIdentifier(dbname), dto);
+            return null;
```
zhljen: For this normalizeIdentifier() function, what if the dbname is an empty string? I guess it will pass into the metacat controller and finally we will catch it once reach the service layer. Wondering if we should catch it here?  
```
if (s == null) {
            return null;
        } else {
            return s.trim().toLowerCase(Locale.ENGLISH);
        }
```

metacat-thrift/src/main/java/com/netflix/metacat/thrift/CatalogThriftHiveMetastore.java
```diff
@@ -976,7 +1006,29 @@ public GetOpenTxnsInfoResponse get_open_txns_info() throws TException {
...
+                partitions.add(hiveConverters.metacatToHivePartition(partition, tableDto));
+            }
+
+            final PartitionSpec pSpec = new PartitionSpec();
+            pSpec.setPartitionList(new PartitionListComposingSpec(partitions));
+            pSpec.setDbName(dbName);
+            pSpec.setTableName(tblName);
+            if (tableDto != null && tableDto.getSerde() != null) {
+                pSpec.setRootPath(tableDto.getSerde().getUri());
+            }
```
zhljen: I am not sure if this is correct. The partition uri can be different from the table serde uri.  If just for information purpose it's fine, I guess.

metacat-thrift/src/main/java/com/netflix/metacat/thrift/CatalogThriftHiveMetastore.java
```diff
@@ -1115,7 +1156,21 @@ public Partition get_partition_with_auth(
...
+                }
+                if (req.getExpr() == null || filter != null) {
+                    //TODO: We need to handle the case for 'hasUnknownPartitions'
+                    final List<Partition> partitions =
+                        getPartitionsByFilter(req.getDbName(), req.getTblName(), filter, req.getMaxParts());
+                    return new PartitionsByExprResult(partitions, false);
+                } else {
+                    throw new MetaException("Failed to deserialize expression - ExprNodeDesc not present");
+                }
+            });
```
zhljen: These set of conditions are very confusing. If the exception is from the ```Utilities.deserializeExpressionFromKryo```, it should be after the first if () like
```
if (req.getExpr() != null) {
           filter = Utilities.deserializeExpressionFromKryo(req.getExpr()).getExprString();
          if ( filter == null ) {
                 throw new MetaException("Failed to deserialize expression - ExprNodeDesc not present");
          }
}
```
Then the second if is the else of if (req.getExpr() != null)
```
if (req.getExpr() != null) {
xxxx
}
else {
                    final List<Partition> partitions =
                      getPartitionsByFilter(req.getDbName(), req.getTblName(), filter, req.getMaxParts());
                    return new PartitionsByExprResult(partitions, false);
}
```

metacat-functional-tests/src/functionalTest/groovy/com/netflix/metacat/MetacatSmokeThriftSpec.groovy
```diff
@@ -148,6 +154,57 @@ class MetacatSmokeThriftSpec extends Specification {
...
+        def invalidDatabaseName = 'test_db1_invalid' + catalogName
+        def databaseName = 'test_db1_' + catalogName
+        client.createDatabase(new Database(databaseName, 'test_db1', null, null))
+        when:
+        client.dropDatabase(databaseName)
+        then:
+        client.getDatabase(databaseName) == null
+        when:
+        client.dropDatabase(invalidDatabaseName)
+        then:
```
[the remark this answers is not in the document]
ajoymajumdar: This is due to the thrift contract where it throws metastore specific exceptions.

metacat-functional-tests/src/functionalTest/groovy/com/netflix/metacat/MetacatThriftFunctionalSpec.groovy
```diff
@@ -457,8 +455,8 @@ class MetacatThriftFunctionalSpec extends Specification {
         db.parameters['field_added'] = now.toString()
         thrift.alterDatabase(db.name, db)
 
-        then: 'TODO remove this once metacat implements alter_database and uncomment the following lines'
-        thrown(Exception)
+        then:
```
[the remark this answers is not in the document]
ajoymajumdar: Will do.

metacat-main/src/main/java/com/netflix/metacat/main/api/v1/MetacatController.java
ajoymajumdar: Will do.

metacat-thrift/src/main/java/com/netflix/metacat/thrift/CatalogThriftHiveMetastore.java
ajoymajumdar: Will do.

metacat-thrift/src/main/java/com/netflix/metacat/thrift/CatalogThriftHiveMetastore.java
ajoymajumdar: Will do.

metacat-thrift/src/main/java/com/netflix/metacat/thrift/CatalogThriftHiveMetastore.java
ajoymajumdar: This method does not validate the name but just changes the case. The validation happens else where. Probably we could move up the validation in here. This will be a separate change.

metacat-thrift/src/main/java/com/netflix/metacat/thrift/CatalogThriftHiveMetastore.java
ajoymajumdar: I have taken this logic from Hive Metastore code.

metacat-thrift/src/main/java/com/netflix/metacat/thrift/CatalogThriftHiveMetastore.java
ajoymajumdar: Done.

PR #262, opened 2018-05-31
On Hive table rename, convert a managed table to an external table.

ajoymajumdar:


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorTableService.java
```diff
@@ -71,6 +71,7 @@
     private final HiveConnectorInfoConverter hiveMetacatConverters;
     private final HiveConnectorDatabaseService hiveConnectorDatabaseService;
     private final boolean allowRenameTable;
+    private final boolean onRenameConvertToExternal;
```
zhljen: Understand that having an option is good, but we have to maintain the option in the config or given a default value. Why don't we force the conversation during rename. I can't think of any reason to set this option as False. One case can be let the client to pass in the parameter to choose convert or not, but this option is global level, it can't do it.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorTableService.java
```diff
@@ -423,6 +428,23 @@ public void rename(
...
+                //
+                final Table table = metacatHiveClient.getTableByName(oldName.getDatabaseName(), oldName.getTableName());
+                Map<String, String> parameters = table.getParameters();
+                if (parameters == null) {
+                    parameters = Maps.newHashMap();
+                    table.setParameters(parameters);
+                }
+                if (!parameters.containsKey(PARAMETER_EXTERNAL)
+                    || parameters.get(PARAMETER_EXTERNAL).equalsIgnoreCase("FALSE")) {
+                    parameters.put(PARAMETER_EXTERNAL, "TRUE");
```
zhljen: You may want to log the conversion here.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorTableService.java
ajoymajumdar: By default, the behavior is always not to move data. If a system wants to give the flexibility to the client, then they will set it to false. Clients will be able to move the data on move. If they don't want to, then they will first set it the table to external and then rename it.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorTableService.java
ajoymajumdar: Keeping it up describes the logic for the whole if block.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorTableService.java
zhljen: Sure, I guess we need to use the default as True always, given kragle can't make the change to V1 for setting the external parameter.

PR #265, opened 2018-06-06
Fixed issue where thrift methods returning partitions were returning partitions with partition values escaped.

ajoymajumdar:
Hive partition names are escaped where as the partition values contain the original value.

### DIFF
--- a/metacat-thrift/src/main/java/com/netflix/metacat/thrift/HiveConvertersImpl.java
+++ b/metacat-thrift/src/main/java/com/netflix/metacat/thrift/HiveConvertersImpl.java
@@ -17,9 +17,7 @@
 package com.netflix.metacat.thrift;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Splitter;
 import com.google.common.base.Strings;
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.netflix.metacat.common.QualifiedName;
@@ -30,6 +28,7 @@ import com.netflix.metacat.common.dto.PartitionDto;
 import com.netflix.metacat.common.dto.StorageDto;
 import com.netflix.metacat.common.dto.TableDto;
 import com.netflix.metacat.common.dto.ViewDto;
+import org.apache.hadoop.hive.common.FileUtils;
 import org.apache.hadoop.hive.metastore.TableType;
 import org.apache.hadoop.hive.metastore.Warehouse;
 import org.apache.hadoop.hive.metastore.api.Database;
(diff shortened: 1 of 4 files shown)


metacat-thrift/src/main/java/com/netflix/metacat/thrift/HiveConvertersImpl.java
```diff
@@ -332,15 +329,19 @@ public PartitionDto hiveToMetacatPartition(final TableDto tableDto, final Partit
             throw new IllegalArgumentException("Invalid partition name", e);
         }
 
-        final List<String> partVals = Lists.newArrayList();
-        for (String key : tableDto.getPartition_keys()) {
-            final String val = hm.get(key);
-            if (val == null) {
-                throw new IllegalArgumentException("Invalid partition name - missing " + key);
+        if (tableDto != null && tableDto.getPartition_keys() != null) {
```
tgianos: `getPartition_keys()` is pretty far from standard java POJO naming convention, is there a reason for this?

Also since you call `getPartition_keys()` here and below why not write it to a local variable first outside the loop that way if the `get` is doing any kind of "expensive" computation you only have to do it once. Also if you do that you can allocate your `partVals` array list with a fixed size

ajoymajumdar: Probably the name was unintentionally copied from the legacy TableDto class. Good point on the get call. Will change it.

metacat-thrift/src/main/java/com/netflix/metacat/thrift/HiveConvertersImpl.java
```diff
@@ -352,18 +358,7 @@ public String getNameFromPartVals(final TableDto tableDto, final List<String> pa
...
-            if (builder.length() > 0) {
-                builder.append('/');
-            }
-
-            builder.append(partitionKeys.get(i));
-            builder.append('=');
-            builder.append(partVals.get(i));
-        }
-        return builder.toString();
+        return FileUtils.makePartName(partitionKeys, partVals, "");
```
tgianos: I'd make this empty string a static constant so you don't have to create a new String on the heap every time.

metacat-thrift/src/main/java/com/netflix/metacat/thrift/HiveConvertersImpl.java
```diff
@@ -331,16 +328,25 @@ public PartitionDto hiveToMetacatPartition(final TableDto tableDto, final Partit
...
-                throw new IllegalArgumentException("Invalid partition name - missing " + key);
+        // Get the partition keys.
+        List<String> partitionKeys = null;
+        if (tableDto != null) {
+            partitionKeys = tableDto.getPartition_keys();
+        }
+        // If table has not been provided, return the values without validating.
+        if (partitionKeys != null) {
+            final List<String> partVals = Lists.newArrayListWithCapacity(partitionKeys.size());
+            for (String key : partitionKeys) {
```
zhljen: Since we allow the non-validation case, Why do we need the validation? The reason to ask is how come we can get partitions that invalid into the system. If that's not possible, we don't need to worry about this.

ajoymajumdar: I did not remove it since the logic was there already. We need to revisit it when we make the second part of the changes.

PR #269, opened 2018-06-18
Throwing internal server error for empty JDO pool exception.

zhljen:


ajoymajumdar:
We should be handling the exception at HMSHandlerProxy and then catching exception, instead of TException, in the connector service layer and throw new ConnectorException.

zhljen:
I am not sure if I understand. The HMSHandlerProxy should not handle this exception specifically, for "get_table()" call, where the exception was thrown from the hivemetastore call as MetaException with the JDOException nested.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorTableService.java
```diff
@@ -117,6 +118,9 @@ public TableInfo get(final ConnectorRequestContext requestContext, final Qualifi
         } catch (NoSuchObjectException exception) {
             throw new TableNotFoundException(name, exception);
         } catch (MetaException exception) {
+            if (exception.getMessage() != null && exception.getMessage().contains("Timeout: Pool empty")) {
```
ajoymajumdar: Will this logic work?

zhljen:
Removed the unused class.

metacat-connector-hive/src/test/groovy/com/netflix/metacat/connector/hive/HiveConnectorTableSpec.groovy
```diff
@@ -41,6 +41,7 @@ import spock.lang.Shared
 import spock.lang.Specification
 import spock.lang.Unroll
 
+
```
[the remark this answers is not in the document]
zhljen: Unit test definitely helps. I will add it.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/metastore/HMSHandlerProxy.java
```diff
@@ -56,12 +61,25 @@ private HMSHandlerProxy(final HiveConf hiveConf, final Registry registry) throws
...
 
         final HMSHandlerProxy handler = new HMSHandlerProxy(hiveConf, registry);
         return (IMetacatHMSHandler) Proxy.newProxyInstance(
-                HMSHandlerProxy.class.getClassLoader(),
-                new Class[]{IMetacatHMSHandler.class}, handler);
+            HMSHandlerProxy.class.getClassLoader(),
+            new Class[]{IMetacatHMSHandler.class}, handler);
+    }
+
+    /**
```
ajoymajumdar: Can't we have this logic in the test class?

zhljen:
Removed the newProxy() function.

PR #270, opened 2018-06-26
Setting timeouts for queries reading hive partitions.

ajoymajumdar:


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/configs/HiveConnectorClientConfig.java
```diff
@@ -152,4 +175,13 @@ private HiveConf getDefaultConf() {
...
         return result;
     }
+
+    private int getDataStoreTimeout(final ConnectorContext connectorContext) {
+        int result = DEFAULT_DATASTORE_READ_TIMEOUT;
+        try {
+            result = Integer.parseInt(
+                connectorContext.getConfiguration().get(HiveConfigConstants.JAVAX_JDO_DATASTORETIMEOUT));
+        } catch (final Exception ignored) { }
+        return result;
```
zhljen: What if the passing value is 0 or a very big number? Should we have error check?

ajoymajumdar: Since this is one-time during initialization and also because it is getting defaulted to 60 secs, I ignore it.

PR #273, opened 2018-07-09
Setting timeouts for queries reading/writing hive partitions. 

ajoymajumdar:
This change is needed to control long running queries.

### DIFF
--- a/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/configs/HiveConnectorClientConfig.java
+++ b/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/configs/HiveConnectorClientConfig.java
@@ -40,7 +40,12 @@ import javax.sql.DataSource;
 @Configuration
 @ConditionalOnProperty(value = "useEmbeddedClient", havingValue = "true")
 public class HiveConnectorClientConfig {
-    private static final int DEFAULT_DATASTORE_READ_TIMEOUT = 60000;
+    /** Default Query timeout in milliseconds. */
+    private static final int DEFAULT_DATASTORE_TIMEOUT = 60000;
+    /** Default Query timeout in milliseconds for reads. */
+    private static final int DEFAULT_DATASTORE_READ_TIMEOUT = 120000;
+    /** Default Query timeout in milliseconds for writes. */
+    private static final int DEFAULT_DATASTORE_WRITE_TIMEOUT = 120000;
 
     /**
      * create local hive client.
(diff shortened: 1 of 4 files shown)


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/configs/HiveConnectorClientConfig.java
```diff
@@ -128,30 +133,34 @@ public DataSourceTransactionManager hiveTxManager(
...
      * @return hive JDBC Template
      */
     @Bean
-    public JdbcTemplate hiveJdbcTemplate(
+    public JdbcTemplate hiveReadJdbcTemplate(
+        final ConnectorContext connectorContext,
         @Qualifier("hiveDataSource") final DataSource hiveDataSource) {
-        return new JdbcTemplate(hiveDataSource);
+        final JdbcTemplate result = new JdbcTemplate(hiveDataSource);
+        result.setQueryTimeout(getDataStoreReadTimeout(connectorContext) / 1000);
```
zhljen: Why not set as second to avoid conversion there?

ajoymajumdar: I am reusing the jdo parameters that already existed.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/configs/HiveConnectorClientConfig.java
```diff
@@ -165,7 +174,7 @@ private HiveConf getDefaultConf(
         final int dataStoreTimeout = getDataStoreTimeout(connectorContext);
         result.setInt(HiveConfigConstants.JAVAX_JDO_DATASTORETIMEOUT, dataStoreTimeout);
         result.setInt(HiveConfigConstants.JAVAX_JDO_DATASTOREREADTIMEOUT, dataStoreTimeout);
```
zhljen: I am confused about the datastore_timeout and datastore_readtimeout. I think this place should use the  getDataStoreReadTimeout() function.

ajoymajumdar: Since hive metastore code is used for table/database creates and updates, and the timeout was earlier set to 60000, I used the same. The read and write are more for the read/writes of partitions and tables(deletes). I could have used the default timeout for the write in hive metastore conf but thought will try out the write timeout to be consistent across.

PR #276, opened 2018-07-24
Mapping SQLException of too many connection to TOO_MANY_REQUEST status

zhljen:


ajoymajumdar:
Like other Metacat exceptions (MetacatException), we need to define an Exception specifically for 429. We should have this mapping of Pool exhaust error either in the connectors or in the controller RequestWrapper, instead of in the ExceptionMapper.

ajoymajumdar:
Changes look good. You also want to modify com.netflix.metacat.client.module.MetacatErrorDecoder.

zhljen:
Updated based on suggestion.

PR #283, opened 2018-08-22
Added cache support

ajoymajumdar:
Added cache support for TableInfo. By default, caching is disabled. Configuration property 'metacat.cache.enabled' setting to true will enable it. If no cache manager is set, the default map is used.

### DIFF
--- a/metacat-common/src/main/java/com/netflix/metacat/common/dto/PartitionDto.java
+++ b/metacat-common/src/main/java/com/netflix/metacat/common/dto/PartitionDto.java
@@ -48,11 +48,11 @@ public class PartitionDto extends BaseDto implements HasDataMetadata, HasDefinit
     @ApiModelProperty(value = "audit information about the partition")
     private AuditDto audit;
     // Marked as transient because we serialize it manually, however as a JsonProperty because Jackson does serialize it
-    @ApiModelProperty(value = "metadata attached to this partition")
+    @ApiModelProperty(value = "metadata attached to the physical data")
     @JsonProperty
     private transient ObjectNode dataMetadata;
     // Marked as transient because we serialize it manually, however as a JsonProperty because Jackson does serialize it
-    @ApiModelProperty(value = "metadata attached to the physical data")
+    @ApiModelProperty(value = "metadata attached to this partition")
     @JsonProperty
     private transient ObjectNode definitionMetadata;
     @ApiModelProperty(value = "the name of this entity", required = true)
--- a/metacat-common/src/main/java/com/netflix/metacat/common/dto/PartitionDto.java
+++ b/metacat-common/src/main/java/com/netflix/metacat/common/dto/PartitionDto.java
@@ -48,11 +48,11 @@ public class PartitionDto extends BaseDto implements HasDataMetadata, HasDefinit
     @ApiModelProperty(value = "audit information about the partition")
     private AuditDto audit;
     // Marked as transient because we serialize it manually, however as a JsonProperty because Jackson does serialize it
-    @ApiModelProperty(value = "metadata attached to the physical data")
+    @ApiModelProperty(value = "Physical metadata: metadata about the physical data referred by the partition.")
     @JsonProperty
     private transient ObjectNode dataMetadata;
     // Marked as transient because we serialize it manually, however as a JsonProperty because Jackson does serialize it
-    @ApiModelProperty(value = "metadata attached to this partition")
+    @ApiModelProperty(value = "Logical metadata: metadata about the logical construct of the partition.")
     @JsonProperty
     private transient ObjectNode definitionMetadata;
     @ApiModelProperty(value = "the name of this entity", required = true)
--- a/metacat-main/src/main/java/com/netflix/metacat/main/services/impl/ConnectorTableServiceProxy.java
+++ b/metacat-main/src/main/java/com/netflix/metacat/main/services/impl/ConnectorTableServiceProxy.java
@@ -1,0 +1,229 @@
+/*
+ *  Copyright 2018 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ */
+
+package com.netflix.metacat.main.services.impl;
+
+import com.google.common.collect.Lists;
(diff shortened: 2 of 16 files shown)


metacat-main/src/main/java/com/netflix/metacat/main/services/impl/ConnectorTableServiceProxy.java
```diff
@@ -0,0 +1,229 @@
...
+        final ConnectorRequestContext connectorRequestContext = converterUtil.toConnectorContext(metacatRequestContext);
+        service.create(connectorRequestContext, converterUtil.fromTableDto(tableDto));
+    }
+
+    /**
+     * Calls the connector table service delete method.
+     * @param name table name
+     */
+    @CacheEvict(key = "'table.' + #name")
+    public void delete(final QualifiedName name) {
```
zhljen: How will the deletion failure be handled during @CacheEvict?

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/ConnectorTableServiceProxy.java
```diff
@@ -0,0 +1,229 @@
...
+        final MetacatRequestContext metacatRequestContext = MetacatContextManager.getContext();
+        final ConnectorTableService service = connectorManager.getTableService(oldName);
+
+        try {
+            log.info("Renaming {} {} to {}", isMView ? "view" : "table", oldName, newName);
+            final ConnectorRequestContext connectorRequestContext
+                = converterUtil.toConnectorContext(metacatRequestContext);
+            service.rename(connectorRequestContext, oldName, newName);
+        } catch (UnsupportedOperationException ignored) {
+        }
```
zhljen: All the rename will evict the table from the cache even though it's not successful?

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/ConnectorTableServiceProxy.java
```diff
@@ -0,0 +1,229 @@
...
+                service.update(connectorRequestContext, converterUtil.fromTableDto(tableDto));
+            } catch (UnsupportedOperationException ignored) {
+            }
+        }
+    }
+
+    private boolean isTableInfoProvided(final TableDto tableDto) {
+        boolean result = false;
+        if ((tableDto.getFields() != null && !tableDto.getFields().isEmpty())
+            || tableDto.getSerde() != null
```
zhljen: Why not check tableDto.getSerde().isEmpty() ?

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/ConnectorTableServiceProxy.java
```diff
@@ -0,0 +1,229 @@
...
+            || (tableDto.getMetadata() != null && !tableDto.getMetadata().isEmpty())
+            || tableDto.getAudit() != null) {
+            result = true;
```
zhljen: Why not check tableDto.getAudit().isEmpty() ?

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/ConnectorTableServiceProxy.java
```diff
@@ -0,0 +1,229 @@
...
+        }
+    }
+
+    /**
+     * Calls the connector table service update method.
+     * @param name table name
+     * @param tableDto table object
+     */
+    @CacheEvict(key = "'table.' + #name")
+    public void update(final QualifiedName name, final TableDto tableDto) {
```
zhljen: Update table always evict cache, I thought that we don't want to evict the cache for the metadata only update.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/ConnectorTableServiceProxy.java
```diff
@@ -0,0 +1,229 @@
...
+        return result;
+    }
+
+    /**
+     * Calls the connector table service exists method.
+     * @param name table name
+     * @return true, if the object exists.
+     */
+    public boolean exists(final QualifiedName name) {
+        final MetacatRequestContext metacatRequestContext = MetacatContextManager.getContext();
```
zhljen: should we bring the table to the cache for the exist() call?

ajoymajumdar: In our initial effort, we use cache only during the getTable API. Later on we can expand it to other internal apis. We can add the support to exists call as well.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/ConnectorTableServiceProxy.java
ajoymajumdar: Because tableDto.getSerde() is not a list.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/ConnectorTableServiceProxy.java
ajoymajumdar: Because tableDto.getAudit() is not a list.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/ConnectorTableServiceProxy.java
ajoymajumdar: OK, I will move the conditional check to TableServiceImpl.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/ConnectorTableServiceProxy.java
ajoymajumdar: Yes, it should be fine since it happens only on UnsupportedOperationException.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/ConnectorTableServiceProxy.java
ajoymajumdar: Cache will not evict if there is a failure.

metacat-common/src/main/java/com/netflix/metacat/common/dto/PartitionDto.java
```diff
@@ -48,11 +48,11 @@
     @ApiModelProperty(value = "audit information about the partition")
     private AuditDto audit;
     // Marked as transient because we serialize it manually, however as a JsonProperty because Jackson does serialize it
-    @ApiModelProperty(value = "metadata attached to this partition")
+    @ApiModelProperty(value = "metadata attached to the physical data")
     @JsonProperty
     private transient ObjectNode dataMetadata;
     // Marked as transient because we serialize it manually, however as a JsonProperty because Jackson does serialize it
-    @ApiModelProperty(value = "metadata attached to the physical data")
+    @ApiModelProperty(value = "metadata attached to this partition")
```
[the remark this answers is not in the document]
ajoymajumdar: No, I just swapped the description because it was incorrect. I will add more details.

PR #293, opened 2018-09-14
Fixed hive integration issues.

ajoymajumdar:
1. Throw TApplicationException when partition filtering fails. During partition pruning, hive ignores this error and continues to query the data without applying predicate pushdown to metastore.
2. Fixed issue where renaming a table to a name that already existed fails when updating the user metadata since in Metacat(based on app configuration) we do not drop the metadata.

### DIFF
--- a/metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlTagService.java
+++ b/metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlTagService.java
@@ -217,6 +217,9 @@ public class MySqlTagService implements TagService {
         try {
             final QualifiedName newName = QualifiedName.ofTable(name.getCatalogName(), name.getDatabaseName(),
                 newTableName);
+            jdbcTemplate.update(SQL_UPDATE_TAG_ITEM,
+                new String[]{MysqlUserMetadataService.DELETE_MARKER + newName.toString(), newName.toString()},
+                new int[]{Types.VARCHAR, Types.VARCHAR});
             jdbcTemplate.update(SQL_UPDATE_TAG_ITEM, new String[]{newName.toString(), name.toString()},
                 new int[]{Types.VARCHAR, Types.VARCHAR});
         } catch (Exception e) {
--- a/metacat-thrift/src/main/java/com/netflix/metacat/thrift/CatalogThriftHiveMetastore.java
+++ b/metacat-thrift/src/main/java/com/netflix/metacat/thrift/CatalogThriftHiveMetastore.java
@@ -122,6 +122,7 @@ import org.apache.hadoop.hive.metastore.api.Type;
 import org.apache.hadoop.hive.metastore.api.UnlockRequest;
 import org.apache.hadoop.hive.metastore.partition.spec.PartitionSpecProxy;
 import org.apache.hadoop.hive.ql.exec.Utilities;
+import org.apache.thrift.TApplicationException;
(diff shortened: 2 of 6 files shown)


metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlTagService.java
```diff
@@ -217,6 +217,9 @@ public void renameTableTags(final QualifiedName name, final String newTableName)
         try {
             final QualifiedName newName = QualifiedName.ofTable(name.getCatalogName(), name.getDatabaseName(),
                 newTableName);
+            jdbcTemplate.update(SQL_UPDATE_TAG_ITEM,
+                new String[]{MysqlUserMetadataService.DELETE_MARKER + newName.toString(), newName.toString()},
+                new int[]{Types.VARCHAR, Types.VARCHAR});
             jdbcTemplate.update(SQL_UPDATE_TAG_ITEM, new String[]{newName.toString(), name.toString()},
```
zhljen: Is this a new way to soft delete? :)

metacat-thrift/src/main/java/com/netflix/metacat/thrift/CatalogThriftHiveMetastore.java
```diff
@@ -1162,17 +1163,26 @@ public Partition get_partition_with_auth(
...
+                    //TODO: We need to handle the case for 'hasUnknownPartitions'
+                    return new PartitionsByExprResult(
+                        getPartitionsByFilter(req.getDbName(), req.getTblName(), filter, req.getMaxParts()),
+                        false);
+                } catch (Exception e) {
+                    //
+                    // Added this to specially handle filter expression failures.
+                    // Extending this to any exception because Hive client then ignores the exception and executes the
+                    // query without predicate pruning.
+                    //
```
zhljen: are you saying the hive client from the thrift side? not sure I get it, since I thought hive client isn't in the loop after we roll out.

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlTagService.java
ajoymajumdar: No. The issue is that when a table ABC is dropped, defintion_metadata for the table is not deleted. When we rename a table XYZ to ABC, the operation fails because there is an existing entry in defintion_metadata.
Now to fix this, I can drop the existing entry and then rename the table. Instead I rename the entry. We should revisit when we implement soft-delete.

metacat-thrift/src/main/java/com/netflix/metacat/thrift/CatalogThriftHiveMetastore.java
ajoymajumdar: We see this issue when running hive queries. Hive client instead of talking to RDS or HCAT directly, it can talk to Metacat.

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlTagService.java
zhljen: that's what I thought. Will our cleanup process pick them up later?

ajoymajumdar: yes.

PR #296, opened 2018-09-28
Fixed issue where the thrift API to rename an iceberg table was failing.

ajoymajumdar:
The issue was that the iceberg table update was failing the validation since the thrift rename API did a rename and update call. The second call to update failed since it was missing the previous_metadata_location. Modified to pass the validation if there is no change in the metadata_location.

### DIFF
--- a/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastTableService.java
+++ b/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastTableService.java
@@ -129,20 +129,25 @@ public class HiveConnectorFastTableService extends HiveConnectorTableService {
             ? existingMetadata.get(DirectSqlTable.PARAM_METADATA_LOCATION) : null;
         final String previousMetadataLocation = newMetadata != null
             ? newMetadata.get(DirectSqlTable.PARAM_PREVIOUS_METADATA_LOCATION) : null;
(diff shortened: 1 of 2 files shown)


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastTableService.java
```diff
@@ -129,20 +129,25 @@ private void validateIcebergUpdate(final TableInfo existingTableInfo, final Tabl
...
-            final String message =
-                "Invalid iceberg table metadata location. Provided previous metadata location is empty.";
-            log.error(message);
-            throw new IllegalStateException(message);
-        } else if (!Objects.equals(existingMetadataLocation, previousMetadataLocation)) {
-            final String message = String.format("Invalid iceberg table metadata location (expected:%s, provided:%s)",
-                existingMetadataLocation, previousMetadataLocation);
-            log.error(message);
-            throw new IllegalStateException(message);
+        } else if (!Objects.equals(existingMetadataLocation, newMetadataLocation)) {
```
zhljen: if the newMetadataLocation is null, it will not check the previous metadata location. Is it ok?

ajoymajumdar: It will since we are checking the inequality.

PR #298, opened 2018-10-04
Supporting table iceberg get partitions

zhljen:


metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/model/PartitionInfo.java
```diff
@@ -53,9 +58,17 @@ private PartitionInfo(
...
+        final StorageInfo serde,
+        final ObjectNode dataMetrics
     ) {
         super(name, auditInfo, metadata);
         this.serde = serde;
+        this.dataMetrics = dataMetrics;
+    }
+
+    @Override
+    public int compareTo(@NonNull @Nonnull final Object o) {
```
ajoymajumdar: We can add this to the BaseInfo.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorPartitionService.java
```diff
@@ -105,26 +108,16 @@ public HiveConnectorPartitionService(
...
-            }
-            return partitionInfos;
-        } catch (NoSuchObjectException exception) {
-            throw new TableNotFoundException(tableName, exception);
-        } catch (MetaException | InvalidObjectException e) {
-            throw new InvalidMetaException("Invalid metadata for " + tableName, e);
-        } catch (TException e) {
-            throw new ConnectorException(String.format("Failed get partitions for hive table %s", tableName), e);
+        final PartitionListRequest partitionsRequest,
+        final TableInfo tableInfo) {
```
ajoymajumdar: I thought we introduced the TableInfo because it was not being called in getPartitions. I see that we used to call. Do we still need to make the wide-spread interface change?

metacat-thrift/src/main/java/com/netflix/metacat/thrift/HiveConvertersImpl.java
```diff
@@ -282,6 +283,9 @@ private StorageDescriptor fromStorageDto(@Nullable final StorageDto storageDto,
         // Setting an empty skewed info.
         result.setSkewedInfo(new SkewedInfo(Collections.emptyList(), Collections.emptyList(), Collections.emptyMap()));
         result.setParameters(sdParams);
+        result.setSkewedInfo(new SkewedInfo(new ArrayList<String>(),
```
ajoymajumdar: We can remove this. We are already doing this above.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/PartitionServiceImpl.java
```diff
@@ -546,4 +562,14 @@ public PartitionDto get(final QualifiedName name) {
     public boolean exists(final QualifiedName name) {
         return get(name) != null;
     }
+
+    private TableInfo getTableInfo(final QualifiedName name) {
```
ajoymajumdar: We can as well pass TableDto instead of TableInfo? Also you can use the Optional method to return.

return tableService.get(name, GetTableServiceParameters.builder().includeInfo(true).useCache(true).build()).orElseThrow(..)

metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/model/PartitionInfo.java
zhljen: Can't move in BaseInfo since we need to cast the class to get Name. I will change the compareTo as an outside comparator.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorPartitionService.java
```diff
@@ -78,24 +78,27 @@
...
+     * @param metacatHiveClient         hive client
+     * @param hiveMetacatConverters     hive converter
+     * @param hiveConnectorTableService hive table service
      */
     public HiveConnectorPartitionService(
         final ConnectorContext context,
         final IMetacatHiveClient metacatHiveClient,
-        final HiveConnectorInfoConverter hiveMetacatConverters
+        final HiveConnectorInfoConverter hiveMetacatConverters,
+        final HiveConnectorTableService hiveConnectorTableService
```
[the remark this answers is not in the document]
zhljen: This was old implementation and will remove.

metacat-thrift/src/main/java/com/netflix/metacat/thrift/HiveConvertersImpl.java
zhljen: Will do.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/PartitionServiceImpl.java
zhljen: Don't want to leak the TableDto into connector implementation. Will use Optional method.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorPartitionService.java
```diff
@@ -105,26 +104,16 @@ public HiveConnectorPartitionService(
...
-            for (Partition partition : partitions) {
-                partitionInfos.add(hiveMetacatConverters.toPartitionInfo(tableInfo, partition));
-            }
-            return partitionInfos;
-        } catch (NoSuchObjectException exception) {
-            throw new TableNotFoundException(tableName, exception);
-        } catch (MetaException | InvalidObjectException e) {
-            throw new InvalidMetaException("Invalid metadata for " + tableName, e);
-        } catch (TException e) {
-            throw new ConnectorException(String.format("Failed get partitions for hive table %s", tableName), e);
```
ajoymajumdar: What is the reason to ignore this exception?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/util/HiveTableUtil.java
```diff
@@ -97,4 +100,26 @@ private static Deserializer createDeserializer(final Class<? extends Deserialize
...
+    }
+
+    /**
+     * get iceberg table metadata location.
+     *
+     * @param tableInfo table info
+     * @return true for iceberg table
+     */
+    public static String getIcebergTableMetadataLocation(final TableInfo tableInfo) {
+        return tableInfo.getMetadata().get(DirectSqlTable.PARAM_METADATA_LOCATION);
```
ajoymajumdar: Are we assuming that this will be called only if it is an iceberg table?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableUtil.java
```diff
@@ -0,0 +1,196 @@
...
+    /**
+     * get data metadata from partition metrics.
+     *
+     * @param metrics metrics.
+     * @return object node of the metrics
+     */
+    public ObjectNode getDataMetadataFromIcebergMetrics(
+        final ScanSummary.PartitionMetrics metrics) {
+        final ObjectNode root = JsonNodeFactory.instance.objectNode();
+        root.set(DataMetricConstants.DATA_METADATA_METRIC_NAME, getMetricValueNode(metrics));
```
ajoymajumdar: This is fine for now. We should have used the dse-metric library to set the values, instead of manipulating the JSON.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableUtil.java
```diff
@@ -0,0 +1,196 @@
...
+
+        IcebergTableOps(final String location) {
+            super(conf);
+            this.location = location;
+            refresh();
+        }
+
+        @Override
+        public TableMetadata refresh() {
+            refreshFromMetadataLocation(this.location);
```
ajoymajumdar: What does this method do?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorPartitionService.java
zhljen: These exceptions are throw from the metacatHiveClient.getTableByName() call, which won't be needed after we passed in the tableInfo.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableUtil.java
zhljen: is dse-metric library open sourced? we may want to stay like this if not.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableUtil.java
zhljen: This is called from iceberg runtime to load the iceberg table.

PR #305, opened 2018-11-08
Upgraded iceberg versions to 0.4.0.

ajoymajumdar:


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableUtil.java
```diff
@@ -80,7 +80,8 @@ public IcebergTableUtil(final ConnectorContext connectorContext) {
                 final Expression filter = (Expression) new PartitionParser(
                     new StringReader(partitionsRequest.getFilter())).filter()
                     .jjtAccept(icebergFilterGenerator, null);
-                return ScanSummary.of(icebergTable.newScan().filter(filter)).build();
+                return ScanSummary.of(icebergTable.newScan().filter(filter))
+                    .limit(connectorContext.getConfig().getIcebergTableSummaryFetchSize()).build();
             }
```
zhljen: Why we add the limit after applying the filter? The filter-based query should return all the results, not the top n results.

ajoymajumdar: Keeping it consistent with the logic without filter. Even with the filter, the results can be huge and slow down the request.

ajoymajumdar: OK. I will remove the limit.

PR #310, opened 2018-11-19
Adding type parameter to tag list.

zhljen:
Adding QualifiedName type parameter to list tag API.

### DIFF
--- a/metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
+++ b/metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
@@ -26,6 +26,7 @@ import lombok.NonNull;
 
 import javax.annotation.Nullable;
 import java.io.Serializable;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
@@ -557,26 +558,53 @@ public final class QualifiedName implements Serializable {
         /**
          * Catalog type.
          */
-        CATALOG,
+        CATALOG("catalog"),
 
         /**
          * Database type.
          */
-        DATABASE,
+        DATABASE("database"),
 
         /**
          * Table type.
          */
-        TABLE,
+        TABLE("table"),
 
         /**
          * Partition type.
          */
-        PARTITION,
+        PARTITION("partition"),
 
         /**
          * MView type.
          */
-        MVIEW
+        MVIEW("mview");
+
+        private final String value;
+
+        /**
+         * Constructor.
+         *
+         * @param value category value.
+         */
+        Type(final String value) {
+            this.value = value;
+        }
+
+        /**
+         * Type create from value.
+         *
+         * @param value string value
+         * @return Type object
+         */
+        public static Type fromValue(final String value) {
+            for (Type type : values()) {
+                if (type.value.equalsIgnoreCase(value)) {
+                    return type;
+                }
+            }
+            throw new IllegalArgumentException(
+                "Unknown enum type " + value + ", Allowed values are " + Arrays.toString(values()));
+        }
     }
 }
--- a/metacat-main/src/main/java/com/netflix/metacat/main/api/v1/TagController.java
+++ b/metacat-main/src/main/java/com/netflix/metacat/main/api/v1/TagController.java
@@ -45,6 +45,8 @@ import lombok.NonNull;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
+import org.springframework.web.bind.WebDataBinder;
+import org.springframework.web.bind.annotation.InitBinder;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
@@ -54,6 +56,7 @@ import org.springframework.web.bind.annotation.ResponseStatus;
 import org.springframework.web.bind.annotation.RestController;
 
 import javax.annotation.Nullable;
+import java.beans.PropertyEditorSupport;
 import java.net.HttpURLConnection;
 import java.util.HashSet;
 import java.util.List;
@@ -143,6 +146,7 @@ public class TagController {
      * @param sourceName   Prefix of the source name
      * @param databaseName Prefix of the database name
      * @param tableName    Prefix of the table name
+     * @param type         metacat qualifed name type
      * @return list of qualified names
      */
     @RequestMapping(
@@ -167,11 +171,20 @@ public class TagController {
         @ApiParam(value = "Prefix of the database name")
         @Nullable @RequestParam(name = "databaseName", required = false) final String databaseName,
         @ApiParam(value = "Prefix of the table name")
-        @Nullable @RequestParam(name = "tableName", required = false) final String tableName
+        @Nullable @RequestParam(name = "tableName", required = false) final String tableName,
+        @ApiParam(value = "Qualified name type")
+        @Nullable
+        @RequestParam(name = "type", required = false) final QualifiedName.Type type
     ) {
         return this.requestWrapper.processRequest(
             "TagV1Resource.list",
-            () -> this.tagService.list(includeTags, excludeTags, sourceName, databaseName, tableName)
+            () -> this.tagService.list(
+                includeTags,
+                excludeTags,
+                sourceName,
+                databaseName,
+                tableName,
+                type)
         );
     }
 
@@ -254,7 +267,7 @@ public class TagController {
             case CATALOG:
                 //catalog service will throw exception if not found
                 final CatalogDto catalogDto = this.catalogService.get(name, GetCatalogServiceParameters.builder()
-                        .includeDatabaseNames(false).includeUserMetadata(false).build());
+                    .includeDatabaseNames(false).includeUserMetadata(false).build());
                 if (catalogDto != null) {
                     return this.tagService.setTags(name, tags, true);
                 }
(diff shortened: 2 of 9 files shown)


metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
```diff
@@ -557,26 +558,53 @@ public String toString() {
         /**
          * Catalog type.
          */
-        CATALOG,
+        CATALOG("catalog"),
```
ajoymajumdar: Why cant we just use UPPERCASE value?

metacat-main/src/main/java/com/netflix/metacat/main/api/v1/TagController.java
```diff
@@ -167,11 +171,20 @@ public TagController(
         @ApiParam(value = "Prefix of the database name")
         @Nullable @RequestParam(name = "databaseName", required = false) final String databaseName,
         @ApiParam(value = "Prefix of the table name")
-        @Nullable @RequestParam(name = "tableName", required = false) final String tableName
+        @Nullable @RequestParam(name = "tableName", required = false) final String tableName,
+        @ApiParam(value = "Qualified name type")
+        @Nullable
+        @RequestParam(name = "type", required = false) final QualifiedName.Type type
```
ajoymajumdar: Why is the type different from that in the API interface?

metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
zhljen: sure. we can change to upcase

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlTagService.java
```diff
@@ -48,11 +50,26 @@
...
+ * @author zhenl
  */
 @Slf4j
 @SuppressFBWarnings
 @Transactional("metadataTxManager")
 public class MySqlTagService implements TagService {
+    /**
+     * QUALIFIEDNAME TYPE REGEX MAP.
+     */
+    private static final Map<QualifiedName.Type, String>
```
ajoymajumdar: Can we add this to the enum?

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlTagService.java
```diff
@@ -307,25 +325,31 @@ public void remove(final QualifiedName name, final Set<String> tags, final boole
...
         //Includes
         final Set<String> localIncludes = includeTags != null ? includeTags : Sets.newHashSet();
         try {
-            String query
-                = String.format(QUERY_SEARCH, "in ('" + Joiner.on("','").skipNulls().join(localIncludes) + "')");
+            StringBuilder query = new StringBuilder(
+                String.format(QUERY_SEARCH, "in ('" + Joiner.on("','").skipNulls().join(localIncludes) + "')"));
             final Object[] params = {localIncludes.size() == 0 ? 1 : 0, wildCardName == null ? 1 : 0, wildCardName};
-            includedNames.addAll(jdbcTemplate.query(query, params,
+            addTypeLimitToTagQuery(query, type);
```
ajoymajumdar: We should add it to the query as a parameter like 1=? or name rlike ?

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlTagService.java
zhljen: why add this mysql regex query to the enum?

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlTagService.java
zhljen: I feel it's hard to understand the logic of "a parameter like 1=?" which isn't needed,  and prefer making it explicitly to add the query if needed.

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlTagService.java
ajoymajumdar: Enum will carry extra information that can be used by any process.

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlTagService.java
ajoymajumdar: Utilizing PreparedStatement approach correctly.

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlTagService.java
```diff
@@ -307,25 +325,31 @@ public void remove(final QualifiedName name, final Set<String> tags, final boole
...
         @Nullable final String databaseName,
-        @Nullable final String tableName
+        @Nullable final String tableName,
+        @Nullable final QualifiedName.Type type
     ) {
         Set<String> includedNames = Sets.newHashSet();
         final Set<String> excludedNames = Sets.newHashSet();
-        final String wildCardName = QualifiedName.toWildCardString(sourceName, databaseName, tableName);
+        final String wildCardName =
+            MySqlServiceUtil.qualifiedNameToWildCardQueryString(sourceName, databaseName, tableName);
```
ajoymajumdar: Why did we replace the previous line with the new one?

metacat-main/src/main/java/com/netflix/metacat/main/api/v1/TagController.java
zhljen: Will change the client to use Type

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlTagService.java
zhljen: Sure.

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlTagService.java
zhljen: Sounds right.

metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
```diff
@@ -550,33 +593,88 @@ public String toString() {
...
          * Table type.
          */
-        TABLE,
+        TABLE("TABLE"),
 
         /**
          * Partition type.
          */
-        PARTITION,
+        PARTITION("PARTITION"),
```
ajoymajumdar: By default, Enum.name() will return the string value as it is defined in the enum. So we do not have to explicitly set it.

metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
```diff
@@ -550,33 +593,88 @@ public String toString() {
...
          * MView type.
          */
-        MVIEW
+        MVIEW("MVIEW");
+
+
+        /**
+         * TYPE REGEX MAP.
+         */
+        private static final Map<Type, String>
```
ajoymajumdar: Instead of using the Map, set the mysqlRegexValue during initialization of an enum. Ex: CATALOG(""^([^\\/]+)$")

metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
```diff
@@ -550,33 +593,88 @@ public String toString() {
...
+            return REGEX_MAP;
+        }
+
+        /**
+         * Type create from value.
+         *
+         * @param value string value
+         * @return Type object
+         */
+        public static Type fromValue(final String value) {
```
ajoymajumdar: We do not need this if we use valueOf

zhljen: To compare the value with equalsIgnoreCase, also giving more information during the exception.

metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
zhljen: This calls constructor with value "PARTITION"., which is used in the converter  in tagController. 
        public void setAsText(final String text) throws IllegalArgumentException {
            super.setValue(QualifiedName.Type.fromValue(text));
        }

metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
zhljen: This looks odd to me, binding the regex into individual enum value.

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlTagService.java
zhljen: It has a similar issue, i.e. only limiting to table and can't return the database using search. From the API signature, it doesn't say the search only limits to tables and we should return the database if it has the tag.

metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
ajoymajumdar: You could use valueOf.

metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
ajoymajumdar: We are binding any context we have about the enum to it. Having it in the Map in this class does not make sense.

zhljen: This regext matching helper doesn't belong to this class, I agree. We can bind to the field if you have a strong opinion, even though I feel it's very strange.

PR #311, opened 2018-11-21
Making Metacat resilient on getPartitions call for tables with high number of partitions.

ajoymajumdar:
GetPartitions will fail if the number of partitions exceeds the configured threshold. This is handled only in hive connector with configurations hive.use.embedded.metastore=true and hive.use.embedded.fastservice=true.

### DIFF
--- a/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableOpWrapper.java
+++ b/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableOpWrapper.java
@@ -45,7 +45,9 @@ public class IcebergTableOpWrapper {
     public Map<String, ScanSummary.PartitionMetrics> getPartitionMetricsMap(final Table icebergTable,
                                                                      @Nullable final Expression filter) {
         return (filter != null)
-            ? ScanSummary.of(icebergTable.newScan().filter(filter)).build()
+            ? ScanSummary.of(icebergTable.newScan().filter(filter))
+                .limit(connectorContext.getConfig().getIcebergTableSummaryFetchSize())
+                .build()
             :
             ScanSummary.of(icebergTable.newScan())  //the top x records
                 .limit(connectorContext.getConfig().getIcebergTableSummaryFetchSize())
--- a/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlGetPartition.java
+++ b/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlGetPartition.java
@@ -39,6 +39,7 @@ import com.netflix.metacat.common.server.partition.parser.PartitionParser;
 import com.netflix.metacat.common.server.partition.util.FilterPartition;
 import com.netflix.metacat.common.server.partition.visitor.PartitionKeyParserEval;
 import com.netflix.metacat.common.server.partition.visitor.PartitionParamParserEval;
+import com.netflix.metacat.common.server.properties.Config;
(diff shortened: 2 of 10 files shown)


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableOpWrapper.java
```diff
@@ -45,7 +45,9 @@
     public Map<String, ScanSummary.PartitionMetrics> getPartitionMetricsMap(final Table icebergTable,
                                                                      @Nullable final Expression filter) {
         return (filter != null)
-            ? ScanSummary.of(icebergTable.newScan().filter(filter)).build()
+            ? ScanSummary.of(icebergTable.newScan().filter(filter))
+                .limit(connectorContext.getConfig().getIcebergTableSummaryFetchSize())
+                .build()
             :
```
zhljen: We should fail the getPartitions call if it exceeds the threshhold instead of return the results within limit.

zhljen: yes, the getPartition call for hive should fail in this case as well.

ajoymajumdar: For iceberg tables, we do not have the control and so we limit it to 100 in the case of no filter. So I added the same logic just for iceberg tables even when a filter is applied. If we want to apply the same in iceberg tables, then it should be done in iceberg api.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlGetPartition.java
```diff
@@ -459,7 +465,20 @@ public int getPartitionCount(
...
+                    final String message =
+                        String.format("Number of partitions queried for table %s exceeded the threshold %d",
+                            tableQName, config.getMaxPartitionsThreshold());
+                    log.warn(message);
+                    throw new IllegalArgumentException(message);
+                }
             }
+            registry.gauge(registry.createId(HiveMetrics.GaugePreExpressionFilterGetPartitionsCount
+                .getMetricName()).withTags(tableQName.parts())).set(noOfRows);
             return result;
```
zhljen: What's the purpose of this metrics? Do we need to know the number of getPartitions before applying the filter? To report the failure case, the above threshhold metric would be useful.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableOpWrapper.java
ajoymajumdar: Are you suggesting we use the same threshold for getPartitions as non-iceberg table and fail if we get back those number of partitions?

zhljen: We should let them know to handle the threshold case.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlGetPartition.java
ajoymajumdar: Yes, this metric will be helpful to monitor the direct sql. This will tell us if we are filtering the partitions on the client side or on the server side.

zhljen: ok.

PR #312, opened 2018-11-27
Modified the check and put logic of iceberg table. 

ajoymajumdar:
The update no longer calls the hive metastore code to alter table. The logic only updates the table parameters in the Hive RDS using direct sqls.

### DIFF
--- a/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
+++ b/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
@@ -205,15 +205,9 @@ public class DirectSqlTable {
             log.warn(message, ex);
             throw new InvalidMetaException(tableName, message, null);
         }
-        if (existingTableMetadata.isEmpty()
-            || !ICEBERG_TABLE_TYPE.equalsIgnoreCase(existingTableMetadata.get(PARAM_TABLE_TYPE))) {
-            final String message = String.format("Originally table %s is not of type iceberg", tableName);
-            log.info(message);
-            throw new InvalidMetaException(tableName, message, null);
-        }
         final Map<String, String> newTableMetadata = tableInfo.getMetadata() == null
             ? Maps.newHashMap() : tableInfo.getMetadata();
-        validateIcebergUpdate(existingTableMetadata, newTableMetadata);
+        validateIcebergUpdate(tableName, existingTableMetadata, newTableMetadata);
         final MapDifference<String, String> diff = Maps.difference(existingTableMetadata, newTableMetadata);
         insertTableParams(tableId, diff.entriesOnlyOnRight());
         final Map<String, String> updateParams = diff.entriesDiffering().entrySet().stream()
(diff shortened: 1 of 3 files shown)


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
```diff
@@ -169,56 +176,93 @@ public boolean exists(final QualifiedName name) {
...
+            || !ICEBERG_TABLE_TYPE.equalsIgnoreCase(existingTableMetadata.get(PARAM_TABLE_TYPE))) {
+            final String message = String.format("Originally table %s is not of type iceberg", tableName);
             log.info(message);
-            throw new IllegalStateException(message);
-        } else {
-            updateTableParam(tableId, PARAM_METADATA_LOCK, 1);
+            throw new InvalidMetaException(tableName, message, null);
         }
+        final Map<String, String> newTableMetadata = tableInfo.getMetadata() == null
+            ? Maps.newHashMap() : tableInfo.getMetadata();
```
zhljen: When will tableInfo.getMetadata() be null? This function is invoked for iceberg table only, which is checked from HiveTableUtil.isIcebergTable(tableInfo). If tableInfo.getMetadata() is null here, something is wrong. We can put a check to ensure not null case or at the very beginning of this function if having the null concern.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
```diff
@@ -169,56 +176,93 @@ public boolean exists(final QualifiedName name) {
...
-            isLocked = jdbcTemplate.queryForObject(SQL.TABLE_PARAM_LOCK,
-                new SqlParameterValue[] {new SqlParameterValue(Types.BIGINT, tableId),
-                    new SqlParameterValue(Types.VARCHAR, PARAM_METADATA_LOCK), }, Boolean.class);
-        } catch (EmptyResultDataAccessException ignored) { }
-        if (isLocked == null) {
-            insertTableParam(tableId, PARAM_METADATA_LOCK, 1);
-        } else if (isLocked) {
-            final String message = String.format("Iceberg table %s is locked.", tableName);
+        if (existingTableMetadata.isEmpty()
+            || !ICEBERG_TABLE_TYPE.equalsIgnoreCase(existingTableMetadata.get(PARAM_TABLE_TYPE))) {
```
zhljen: if (existingTableMetadata.get(PARAM_METADATA_LOCATION) ) check move here as well. It's better to have a function to check/valide the existingTableMetadata .

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
```diff
@@ -169,56 +176,93 @@ public boolean exists(final QualifiedName name) {
...
+                    "Invalid iceberg table metadata location. Provided previous metadata location is empty.";
+                log.error(message);
+                throw new IllegalStateException(message);
+            } else if (!Objects.equals(existingMetadataLocation, previousMetadataLocation)) {
+                final String message =
+                    String.format("Invalid iceberg table metadata location (expected:%s, provided:%s)",
+                        existingMetadataLocation, previousMetadataLocation);
+                log.error(message);
+                throw new IllegalStateException(message);
+            }
```
zhljen: If I understand right, the validation logic is 
1. existingTableMetadata is valid ( existingMetadataLocation  not blank..). 
2. newTableMetadata is valid (previousMetadataLocation  not blank )
3. existingMetadataLocation = previousMetadataLocation
The above code will be cleaner if refactor a bit

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
```diff
@@ -169,56 +176,93 @@ public boolean exists(final QualifiedName name) {
...
         try {
-            tableType = jdbcTemplate.queryForObject(SQL.TABLE_PARAM_LOCK,
-                new SqlParameterValue[] {new SqlParameterValue(Types.BIGINT, tableId),
-                    new SqlParameterValue(Types.VARCHAR, PARAM_TABLE_TYPE), }, String.class);
-        } catch (EmptyResultDataAccessException ignored) { }
-        if (tableType == null || !ICEBERG_TABLE_TYPE.equalsIgnoreCase(tableType)) {
-            final String message = String.format("Originally table %s is not of type iceberg", tableName);
-            log.info(message);
+            existingTableMetadata = jdbcTemplate.query(SQL.TABLE_PARAMS_LOCK,
+                new SqlParameterValue[]{new SqlParameterValue(Types.BIGINT, tableId)}, rs -> {
```
zhljen: Now we are locking all the rows of this table parameters.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
ajoymajumdar: Good point. Since this is a public method is a separate class, I added it as a safety.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
ajoymajumdar: Yes.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
ajoymajumdar: Yes. In this PR, I just moved the code from HIveConnectorFastTableService to this class.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
ajoymajumdar: I will move this check to the validateIcebergUpdate method.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
zhljen: We should fail in this case, otherwise the following validateIcebergUpdate() function will fail with a miss-leading exception message.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
zhljen: Should we make the code better since we are touching them?

ajoymajumdar: I feel it is good. If you have any suggestions, let me know.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
ajoymajumdar: It will fail saying the metadata location is empty. Are you suggesting, fail even before locking the records?

PR #313, opened 2018-11-30
Optimization: Modified to not initialize pig objects that our converter logic does not use

ajoymajumdar:
Earlier code was creating a pig context that created long live objects that our code does not use

### DIFF
--- a/metacat-connector-pig/src/main/java/com/netflix/metacat/connector/pig/converters/PigTypeConverter.java
+++ b/metacat-connector-pig/src/main/java/com/netflix/metacat/connector/pig/converters/PigTypeConverter.java
@@ -31,21 +31,34 @@ import com.netflix.metacat.common.type.Type;
 import com.netflix.metacat.common.type.TypeUtils;
 import com.netflix.metacat.common.type.VarcharType;
 import lombok.NonNull;
+import org.apache.pig.backend.executionengine.ExecutionEngine;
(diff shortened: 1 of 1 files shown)


metacat-connector-pig/src/main/java/com/netflix/metacat/connector/pig/converters/PigTypeConverter.java
```diff
@@ -31,29 +31,43 @@
...
 
     /**
      * {@inheritDoc}.
      */
     @Override
     public Type toMetacatType(@Nonnull @NonNull final String pigType) {
         try {
-            final LogicalSchema schema = Utils.parseSchema(pigType);
+            final LogicalSchema schema = new QueryParserDriver(PIG_CONTEXT,
+                "util", new HashMap<>()).parseSchema(pigType);
```
zhljen: Can we make this QueryParserDriver instance to be reused ( instead of making the PigContext reused )?

ajoymajumdar: I am not sure if it is threadsafe. There is no documentation. I have to look into the method to decide that.

zhljen: sound reasonable to me, if you have identified the pigContext object is the root not this QueryParserDriver object.

PR #316, opened 2018-12-11
Add timeouts to calls to ElasticSearch.

ajoymajumdar:
Add configuration to enable/disable the sync of data metadata for tables with same uri.


metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchUtilImpl.java
```diff
@@ -118,7 +122,7 @@ public ElasticSearchUtilImpl(
     public void delete(final String type, final String id) {
         try {
             RETRY_ES_PUBLISH.call(() -> {
-                client.prepareDelete(esIndex, type, id).execute().actionGet();
+                client.prepareDelete(esIndex, type, id).execute().actionGet(esCallTimeout);
                 return null;
```
zhljen: Can we catch the exception due to timeout() , i.e. ElasticsearchTimeoutException() and metrics them separately? So that we know how the timeout configures perform.

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchUtilImpl.java
```diff
@@ -118,16 +126,53 @@ public ElasticSearchUtilImpl(
...
             });
         } catch (Exception e) {
-            log.error("Failed deleting metadata of type {} with id {}. {}", type, id, e);
-            this.elasticSearchMetric.getElasticSearchDeleteFailureCounter().increment();
-            log("ElasticSearchUtil.delete", type, id, null, e.getMessage(), e, true);
+            handleException("ElasticSearchUtil.delete", type, id, e,
+                Metrics.CounterElasticSearchDelete.getMetricName());
```
zhljen: Great!

metacat-main/src/main/java/com/netflix/metacat/main/services/search/ElasticSearchUtilImpl.java
```diff
@@ -118,16 +126,53 @@ public ElasticSearchUtilImpl(
...
+            final Throwable error = ((RetryException) exception).getLastFailedAttempt().getExceptionCause();
+            if (error != null) {
+                exceptionName = error.getClass().getSimpleName();
+            }
+        }
+        final Map<String, String> tags = ImmutableMap
+            .<String, String>builder().put("status", "failure").put("name", id).put("exception", exceptionName).build();
+        registry.counter(registry.createId(metricName).withTags(tags)).increment();
+        log(request, type, id, null, exception.getMessage(), exception, true);
+    }
```
zhljen: The only difference between these two functions is the 'id' vs 'ids' parameter and how we use them in the tag and log. My suggestion is to rename the second one to handleExceptionForBulkOperation or something similar.

PR #317, opened 2018-12-18
Update to iceberg 0.6.2 and fix for handling timestamp query in iceberg.

zhljen:


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/util/IcebergFilterGenerator.java
```diff
@@ -200,4 +130,121 @@ private Expression evalSingleTerm(final ASTCOMPARE node, final Object data) {
...
+     * Get the key and value field of iceberg expression.
+     *
+     * @param lhs left hand string
+     * @param rhs right hand string
+     * @return key value pair for iceberg expression.
+     */
+    private Pair<String, Object> getExpressionKeyValue(final Object lhs,
+                                                       final Object rhs) {
+        if (isIcebergTimestamp(lhs)) {
+            return new ImmutablePair<>(lhs.toString(), ((BigDecimal) rhs).longValue());
```
ajoymajumdar: Any reason using ImmutablePair instead of Map.Entry?

zhljen: The pair is more appropriate here since it can only have two values left hand value and right hand value.

PR #322, opened 2019-01-17
Updating iceberg to use predictable uri

zhljen:


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastPartitionService.java
```diff
@@ -395,4 +398,18 @@ public void deletePartitions(
     private Date fromEpochMilliToDate(@Nullable final Long l) {
         return (l == null) ? null : Date.from(Instant.ofEpochMilli(l));
     }
+
+    //iceberg://<db-name.table-name>/<partition>/snapshot_time=<dateCreated>
+    private String getIcebergPartitionURI(final String databaseName,
```
ajoymajumdar: We could have made this an interface method and have an internal implementation. For now, lets move on with this.

PR #325, opened 2019-01-25
Upgrade to Spring Boot 2.0.8

raveeram:
- Remove dependency on the Spring Platform BOM.
- Fix a couple of new potential NPEs surfaced by findbugs.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -19,21 +19,20 @@
 ## Spring Dependency Versions
 
 # Used in documentation and for including the Gradle plugin
-spring_boot_version=1.5.9.RELEASE
-spring_cloud_version=Edgware.RELEASE
-spring_platform_version=Brussels-SR6
+spring_boot_version=2.0.8.RELEASE
+spring_cloud_version=Finchley.SR2
+google_guice_version=4.1.0
 
-## Override Spring Platform IO Versions
+## Override Spring Boot versions
 
 elasticsearch.version=5.4.1
 guava.version=19.0
-guice.version=4.0
 jersey.version=2.19
-tomcat.version=8.0.22
 
 ## Versions Not Covered by Spring BOMs
 
 hive_version=1.2.1
+
 ## Gradle Properties
 
 org.gradle.parallel=false
--- a/metacat-client/src/main/java/com/netflix/metacat/client/module/MetacatErrorDecoder.java
+++ b/metacat-client/src/main/java/com/netflix/metacat/client/module/MetacatErrorDecoder.java
@@ -34,6 +34,8 @@ import feign.Util;
 import lombok.AllArgsConstructor;
 
 import java.io.IOException;
+import java.time.Instant;
+import java.util.Date;
 
 /**
  * Module that provides a error decoder, used to parse errors.
@@ -77,7 +79,7 @@ public class MetacatErrorDecoder extends feign.codec.ErrorDecoder.Default {
                     return new MetacatTooManyRequestsException(message);
                 case 500: //INTERNAL_SERVER_ERROR
                 case 503: //SERVICE_UNAVAILABLE
-                    return new RetryableException(message, null);
+                    return new RetryableException(message, null, new Date(Instant.now().toEpochMilli()));
                 default:
                     return new MetacatException(message);
             }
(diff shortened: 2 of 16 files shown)


metacat-client/src/main/java/com/netflix/metacat/client/module/MetacatErrorDecoder.java
```diff
@@ -77,7 +78,7 @@ public Exception decode(final String methodKey, final Response response) {
                     return new MetacatTooManyRequestsException(message);
                 case 500: //INTERNAL_SERVER_ERROR
                 case 503: //SERVICE_UNAVAILABLE
-                    return new RetryableException(message, null);
+                    return new RetryableException(message, null, new Date());
```
ajoymajumdar: Use the java.time.Instant class

metacat-main/src/main/java/com/netflix/metacat/main/configs/MetricsConfig.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * A default registry.
+     *
+     * @return The registry to use.
+     */
+    @Bean
+    @ConditionalOnMissingBean(Registry.class)
+    public Registry spectatorRegistry() {
+        return new DefaultRegistry();
```
twicksell: You won't want this when we pull things internal. I'd suggest maybe setting a `@ConditionalOnProperty` to disable this more explicitly. The missing bean condition is probably going to get evaluated out of order relative to our internal auto configs, so that might not get it done.

metacat-main/src/main/resources/application.yml
```diff
@@ -17,12 +17,13 @@
...
-  contextPath: /actuator
   health:
     elasticsearch:
       enabled: false # This is needed for now as we're using a very out of date ES client that throws runtime errors
-  security:
-    enabled: false
+  endpoints:
+    web:
+      exposure:
+        include: "*"
```
twicksell: This might be a little overly aggressive. There are some endpoints like the shutdown and trace ones which are a little dangerous. We turn those off internally.

metacat-main/src/main/resources/application.yml
```diff
@@ -46,3 +47,5 @@ spring:
     default-property-inclusion: always
     deserialization:
       FAIL_ON_UNKNOWN_PROPERTIES: false
+  main:
+    allow-bean-definition-overriding: true
```
twicksell: If you can avoid needing to set this, that'd be for the best long term. But not sure what specifically caused you to need this.

metacat-main/src/main/resources/application.yml
raveeram: I see. Excluded those (shutdown is disabled by default).

metacat-main/src/main/java/com/netflix/metacat/main/configs/MetricsConfig.java
raveeram: Makes sense. Added a `@ConditionalOnProperty`

metacat-main/src/main/resources/application.yml
raveeram: Yeah, it looks like we unintentionally override beans in a few places. I added this as a short-term fix for Boot 2.1 since it's false by default there. Shouldn't be necessary for Boot 2.0. I'll make sure to fix the overrides when we upgrade to 2.1.

gradle.properties
```diff
@@ -19,21 +19,21 @@
 ## Spring Dependency Versions
 
 # Used in documentation and for including the Gradle plugin
-spring_boot_version=1.5.9.RELEASE
-spring_cloud_version=Edgware.RELEASE
-spring_platform_version=Brussels-SR6
+spring_boot_version=2.0.8.RELEASE
+spring_cloud_version=Finchley.SR2
+google_guice_version=4.0
```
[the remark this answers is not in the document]
raveeram: Shouldn't matter but bumped it up to 4.1 and validated metacat-netflix as well.

metacat-client/src/main/java/com/netflix/metacat/client/module/MetacatErrorDecoder.java
raveeram: Ah this is a Feign API that requires a java.util.Date.

ajoymajumdar: I meant using  >> new Date(Instant.now().toEpochMilli()).

raveeram: Ah I see why, fixed, thanks.

PR #331, opened 2019-02-13
Fix exception handling on Get after UpdateTable

raveeram:
- Previously, we were only handling TableNotFoundExceptions. But looks like we need to handle various other types as well (eg. Iceberg could throw FileNotFoundException). We fix this by catching all exceptions on a Get() after a successful Update(). This is inline with the semantics defined for the UpdateTable API as well.
- Added a functional test specifically to test this scenario end-to-end.

### DIFF
--- a/metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
+++ b/metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
@@ -378,13 +378,18 @@ public class TableServiceImpl implements TableService {
                 .record(duration, TimeUnit.MILLISECONDS);
         }
 
-        final TableDto updatedDto = get(name,
-            GetTableServiceParameters.builder()
-                .disableOnReadMetadataIntercetor(false)
-                .includeInfo(true)
-                .includeDataMetadata(true)
-                .includeDefinitionMetadata(true)
-                .build()).orElse(tableDto);
+        TableDto updatedDto = tableDto;
+        try {
+            updatedDto = get(name,
+                GetTableServiceParameters.builder()
+                    .disableOnReadMetadataIntercetor(false)
+                    .includeInfo(true)
+                    .includeDataMetadata(true)
+                    .includeDefinitionMetadata(true)
+                    .build()).orElse(tableDto);
+        } catch (Exception e) {
+            log.warn("Failed to read updated TableDto", e);
+        }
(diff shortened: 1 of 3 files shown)


metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -378,13 +378,18 @@ public TableDto updateAndReturn(final QualifiedName name, final TableDto tableDt
...
+            updatedDto = get(name,
+                GetTableServiceParameters.builder()
+                    .disableOnReadMetadataIntercetor(false)
+                    .includeInfo(true)
+                    .includeDataMetadata(true)
+                    .includeDefinitionMetadata(true)
+                    .build()).orElse(tableDto);
+        } catch (Exception e) {
+            log.warn("Failed to read updated TableDto", e);
+        }
```
zhljen: This looks like a best effort approach to get the updated table information.  It will work from metacat side, but it will have side effects.  What if the get() operation fails and the updateDto will be same as tableDto. In this case, the updateDto will be the same as tableDto, they will miss one update event. But we don't know if this Get() failure can be ignored or not.   One option is to check all the exception from icegberg side and process them explicitly, option 2 is not read the tableInfo for iceberg table (which can be identified using the tableDto ).

zhljen: At least, we should put comments to explain the context why this change is for and what the impact can be.

raveeram: Ah they will not miss the update event because we send an empty SNS notification when tableDto is equal to updateDto which should prompt the client to retry and call back. But agree, an update to the method comment itself is probably needed. Will update.

zhljen: sure, not sure "which client to call back" means. We should ensure downstream consumers on the same page.

raveeram: By client, I meant the downstream services that depend on SNS event notifications. To be clear, the event is fired regardless of the get() call failing. It's just that the TableDto might not be the latest. This seemed like the right behavior even for non-iceberg scenarios too. We also want to perhaps avoid making iceberg specific changes at the service layer. We can discuss more offline if that works

PR #337, opened 2019-03-28
Modified CatalogTraversal to batch up the processing of catalogs.

ajoymajumdar:


metacat-main/src/main/java/com/netflix/metacat/main/services/CatalogTraversal.java
```diff
@@ -231,6 +218,15 @@ private void shutdown(@Nullable final ExecutorService executorService) {
 
     @SuppressWarnings("checkstyle:methodname")
     private ListenableFuture<Void> _processCatalogs(final List<String> catalogNames) {
+        log.info("Start: Full traversal of catalogs: {}", catalogNames);
+        final List<List<String>> subCatalogNamesList = Lists.partition(catalogNames, 5);
```
raveeram: Do we want to hardcode this?

PR #338, opened 2019-04-02
Modified to ignore exception that happens after a successful iceberg table update.

ajoymajumdar:


metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -367,33 +368,72 @@ public TableDto updateAndReturn(final QualifiedName name, final TableDto tableDt
         // Check if the table schema info is provided. If provided, we should continue calling the update on the table
         // schema. Uri may exist in the serde when updating data metadata for a table.
         //
+        boolean ignoreErrorsAfterUpdate = false;
         if (isTableInfoProvided(tableDto, oldTable)) {
-            connectorTableServiceProxy.update(name, converterUtil.fromTableDto(tableDto));
+            ignoreErrorsAfterUpdate = connectorTableServiceProxy.update(name, converterUtil.fromTableDto(tableDto));
```
raveeram: Would probably be simpler to just check the TableInfo here to see if this is an Iceberg table here instead of passing a context back and forth? Seems a bit overkill :)

PR #340, opened 2019-04-30
Mysql Connector Not Handling Table Names with Upper Case

karthik1289:
Problem Description: Table names which are created in Upper case in MYSQL is not searchable. Problem Cause: Under QualifiedName we are standardizing to lowercase for all database and table names. Its better to leave them as is as relational databases might be case sensitive. Problem Solution: Not standardizing the database, table and view Names


ajoymajumdar:
Early on, we made a decision to lower case names in Metacat. QualifiedName represents the identifier for an object in a catalog, where we standardize the names to be lower-cased.

karthik1289:
mysql is case sensitive. It wont accept search with lowercase for table names with upper case.

karthik1289:
if this is not the right approach then atleast we have to check for harvester is mysql and in that case we shouldn't standardize. Please provide your thoughts on the same.

PR #341, opened 2019-04-30
Fixing NPE in Druid Harvester

karthik1289:
Problem Description: In druid harvester, if loadSpec doesn't contains bucket and keys then we are getting Null Pointer Exception. 
Problem Solution: DruidConverterUtil doesn't check for null for the bucket and keys in loadSpecNode


karthik1289:
Tested with Druid Imply 2.9.4

raveeram:
Thanks for submitting this fix @karthik1289 !

zhljen:
Looks good to me.

karthik1289:
Could we merge PR then if its approved?

PR #343, opened 2019-05-10
Cleaned up dependencies for metacat-client and metacat-common.

ajoymajumdar:
All transitive dependencies except metacat-common are relocated in the shaded jar.

#1. Upgraded gradle to 5.4.1
#2. Cleaned up unwanted dependencies.
#3. Added functionalTestShaded to test the shaded jar.


zhljen:
Wondering if we should create this shade client as internal metacat, which will be easier to maintain.

ajoymajumdar:
Need to keep the binaries along with its other jar. This way you maintain the version. We are shading all transitive dependencies.

metacat-functional-tests/src/functionalTestShaded/groovy/com/netflix/metacat/MetacatSmokeSpec.groovy
```diff
@@ -0,0 +1,1467 @@
...
+import java.util.concurrent.TimeUnit
+
+/**
+ * MetacatSmokeSpec.
+ * @author amajumdar
+ * @author zhenl
+ * @since 1.0.0
+ */
+class MetacatSmokeSpec extends Specification {
+    public static MetacatV1 api
```
zhljen: Are these tests are the same as the existing functional test except for testing shaded jar?

metacat-functional-tests/build.gradle
```diff
@@ -63,18 +67,20 @@ dependencies {
      *******************************/
 
     // Test that metacat-client can be used with jersey 1
-    testCompile(project(":metacat-client")) {
-        exclude group: "javax.ws.rs", module: "javax.ws.rs-api"
-        exclude group: "org.glassfish.jersey.core"
-    }
+    functionalTestShadedCompile project(path: ':metacat-client', configuration: 'shadow')
+
```
zhljen: I don't understand this 'Test that metacat-client can be used with jersey 1'..

zhljen:
I still wonder if we can consider pulling back the fat client internally, maybe as a separate project.

metacat-client/build.gradle
```diff
@@ -56,3 +67,26 @@ dependencies {
 
     testCompile(project(":metacat-common").sourceSets.test.output)
 }
+
+// All transitive dependencies are relocated except metacat-common.
+shadowJar {
```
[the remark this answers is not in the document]
ajoymajumdar: 5.3MB. non-shaded jar is 19KB

metacat-functional-tests/src/functionalTestShaded/groovy/com/netflix/metacat/MetacatSmokeSpec.groovy
ajoymajumdar: Right.

metacat-functional-tests/build.gradle
ajoymajumdar: This was an old comment. Then the platform supported jersey 1. We can remove the comment and dependency in the test.

danielbwatson: That is really old, from before Metacat moved to Spring.  When we were transitioning to Java 8 I added a test to make sure Metacat worked with both Jersey 1 and 2.  It looks like only the comment remains and can be safely removed.

PR #344, opened 2019-05-21
Added account id to the cluster dto/info classes.

ajoymajumdar:


metacat-common/src/main/java/com/netflix/metacat/common/dto/ClusterDto.java
```diff
@@ -32,12 +32,15 @@
     @ApiModelProperty(value = "the type of the cluster", required = true)
     private String type;
     /** Account under which the cluster exists. Ex: "abc_test" */
-    @ApiModelProperty(value = "the account type for this catalog", required = false)
+    @ApiModelProperty(value = "the account name under which the cluster exists", required = false)
```
raveeram: under which the cluster was created?

ajoymajumdar: Will correct it.

PR #347, opened 2019-05-28
Fixed issue where getPartition calls for audit tables were falling back to default logic after failing query because of invalid SQL

ajoymajumdar:
Also, on sql timeouts, made the code change to propagate the error instead of falling back to the default logic.

### DIFF
--- a/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlGetPartition.java
+++ b/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlGetPartition.java
@@ -50,6 +50,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.hadoop.hive.metastore.api.FieldSchema;
 import org.apache.hadoop.hive.metastore.api.Table;
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.dao.QueryTimeoutException;
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.jdbc.core.ResultSetExtractor;
 import org.springframework.jdbc.core.SqlParameterValue;
@@ -551,6 +552,7 @@ public class DirectSqlGetPartition {
         final boolean forceDisableAudit
     ) {
         List<T> partitions;
+        final QualifiedName tableQName = QualifiedName.ofTable(catalogName, databaseName, tableName);
         try {
             if (!Strings.isNullOrEmpty(filterExpression)) {
                 final PartitionFilterGenerator generator =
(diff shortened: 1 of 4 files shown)


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlGetPartition.java
```diff
@@ -573,6 +574,10 @@ public int getPartitionCount(
                     null, null,
                     null, sort, pageable, forceDisableAudit);
             }
+        } catch (QueryTimeoutException e) {
+            // On query timeouts, propagate the error.
```
zhljen: How long is the timeout? and what will Metacat return back to the client side for indicating this is a timeout?

ajoymajumdar: It is configurable. Like today, it will return a HTTP 500 error.

zhljen: Just from the PR, it's hard for me to trace how the exception is processed through the server to the client. My suggestion is 
- add logging here to indicate the time out happens with table information
- ensure the timeout message is passed to the client so that they can decide either reduce the number of getPartitions or retry

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlGetPartition.java
```diff
@@ -573,6 +574,10 @@ public int getPartitionCount(
...
+            registry.counter(HiveMetrics.CounterHiveGetTablePartitionsTimeoutFailure.getMetricName()).increment();
+            throw e;
```
zhljen: Can we add the table tag into this metric? Not sure how often this happens, and it will help us trace the heavy loaded query/tables.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlGetPartition.java
ajoymajumdar: Will add comment

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlGetPartition.java
```diff
@@ -918,14 +923,17 @@ private String prepareFilterSql(@Nullable final String filterExpression) {
             if (partitionIds != null && !partitionIds.isEmpty()) {
                 paramsBuilder.addAll(partitionIds);
             }
+            if (filterSql != null && filterParams != null) {
```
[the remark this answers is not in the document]
ajoymajumdar: Yes, it is. This is the bug that the code change fixes. Audit query does a union of two similar queries and hence the params are the same for both the queries.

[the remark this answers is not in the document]
raveeram: Ah see it now, this actually fixes the query.

PR #348, opened 2019-05-30
Fixed issue where getTable API fails with a HTTP 500 instead of HTTP 404 when getting table from a snowflake catalog

ajoymajumdar:


metacat-connector-snowflake/src/main/java/com/netflix/metacat/connector/snowflake/SnowflakeConnectorTableService.java
```diff
@@ -152,12 +152,16 @@ protected ResultSet getColumns(
         @Nonnull @NonNull final Connection connection,
         @Nonnull @NonNull final QualifiedName name
     ) throws SQLException {
```
raveeram: This wouldn't be true anymore right? Looks like the caller of this method, get() in JdbcConnectorTableService already catches this. Wonder if this fix to go through the ExceptionMapper should be done there instead to catch similar exceptions from other methods?

metacat-connector-snowflake/src/main/java/com/netflix/metacat/connector/snowflake/SnowflakeConnectorTableService.java
```diff
@@ -152,12 +152,16 @@ protected ResultSet getColumns(
...
+        try {
+            return connection.getMetaData().getColumns(
+                connection.getCatalog(),
+                name.getDatabaseName(),
+                name.getTableName(),
+                JdbcConnectorUtils.MULTI_CHARACTER_SEARCH
+            );
+        } catch (SQLException e) {
+            throw this.exceptionMapper.toConnectorException(e, name);
+        }
```
zhljen: It looks that this is the fix for the getColumns failure with tableNotFound. Another way to do is to check in the method ```public TableInfo get()```.

metacat-connector-snowflake/src/main/java/com/netflix/metacat/connector/snowflake/SnowflakeConnectorTableService.java
ajoymajumdar: True. Your suggestion is right but that was the case before we made a bug fix.

PR #349, opened 2019-06-04
Timeout the get partitions call to iceberg.

ajoymajumdar:
By default the timeout is 120 secs. On timeout, task that makes the iceberg call is cancelled.

### DIFF
--- a/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableOpWrapper.java
+++ b/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableOpWrapper.java
@@ -16,14 +16,23 @@
  */
 package com.netflix.metacat.connector.hive.iceberg;
 
+import com.google.common.base.Throwables;
+import com.google.common.collect.Maps;
 import com.netflix.iceberg.ScanSummary;
 import com.netflix.iceberg.Table;
 import com.netflix.iceberg.expressions.Expression;
 import com.netflix.metacat.common.server.connectors.ConnectorContext;
-import lombok.AllArgsConstructor;
+import com.netflix.metacat.common.server.properties.Config;
+import com.netflix.metacat.common.server.util.ThreadServiceManager;
+import com.netflix.metacat.connector.hive.util.HiveConfigConstants;
+import lombok.extern.slf4j.Slf4j;
 
 import javax.annotation.Nullable;
 import java.util.Map;
+import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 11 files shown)


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableOpWrapper.java
```diff
@@ -44,14 +67,31 @@
...
             :
             ScanSummary.of(icebergTable.newScan())  //the top x records
-                .limit(connectorContext.getConfig().getIcebergTableSummaryFetchSize())
-                .build();
+                .limit(config.getIcebergTableSummaryFetchSize())
+                .build());
+        try {
+            final int getIcebergPartitionsTimeout = Integer.parseInt(configuration
+                .getOrDefault(HiveConfigConstants.GET_ICEBERG_PARTITIONS_TIMEOUT, "120"));
+            result = future.get(getIcebergPartitionsTimeout, TimeUnit.SECONDS);
```
zhljen: Feel 120 seconds is quite short, not sure if how many tables can survive it. Maybe 5 minutes?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlGetPartition.java
```diff
@@ -522,9 +522,20 @@ public int getPartitionCount(
...
                         "serde_id", serdeParams)));
             }
+            ListenableFuture<List<Void>> future = null;
             try {
-                Futures.transform(Futures.successfulAsList(futures), Functions.constant(null)).get(1, TimeUnit.HOURS);
-            } catch (Exception e) {
+                future = Futures.allAsList(futures);
+                final int getPartitionsDetailsTimeout = Integer.parseInt(configuration
+                    .getOrDefault(HiveConfigConstants.GET_PARTITION_DETAILS_TIMEOUT, "120"));
+                future.get(getPartitionsDetailsTimeout, TimeUnit.SECONDS);
```
zhljen: How the 120 seconds was picked? If it's too short, we may get more pressure from the applications retries.  Can we get a rough idea based on the current metrics?

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableOpWrapper.java
ajoymajumdar: 120 secs is the default. We can always configure it.

PR #354, opened 2019-07-18
Fixed issue with renaming a View in hive.

ajoymajumdar:


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/HiveConnectorTableService.java
```diff
@@ -323,9 +323,6 @@ protected void update(final ConnectorRequestContext requestContext,
                           final Table existingTable, final TableInfo tableInfo) {
         final QualifiedName tableName = tableInfo.getName();
         try {
-            if (existingTable.getTableType().equals(TableType.VIRTUAL_VIEW.name())) {
-                throw new TableNotFoundException(tableName);
-            }
             updateTable(requestContext, existingTable, tableInfo);
```
zhljen: Changes look good to me. Just curious why we block view in this function.

ajoymajumdar: Wanted to check with you because you added this.

zhljen: Per commit log, this was added in the hive refractor change which remained the same business logic as before.  I don't know the intention of this line specifically. If you think we should remove, please go ahead.

PR #355, opened 2019-07-23
Update the table location when updating an iceberg table.

ajoymajumdar:


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
```diff
@@ -223,6 +224,11 @@ public void updateIcebergTable(final TableInfo tableInfo) {
         final Map<String, String> updateParams = diff.entriesDiffering().entrySet().stream()
             .collect(Collectors.toMap(Map.Entry::getKey, s -> s.getValue().rightValue()));
         updateTableParams(tableId, updateParams);
+        //
+        // In addition to updating the table params, the table location in HMS needs to be updated for usage by
+        // external tools, that access HMS directly
+        //
+        updateTableLocation(tableId, tableInfo);
```
zhljen: What if this update failed? Will the updateTableParams still succeed? If it's part of contract, we should fail the updateTableParameter. Also, the update location should be part of the validation to ensure the location is valid, and also sync with /metadata/ location ( the assumption from iceberg table).

ajoymajumdar: If the update fails, the whole transaction is not committed. We can add the validation later because I am not sure of the assumption.

zhljen: Not sure what the assumption refers to. I was told that metacat should have the table location as the same as iceberg table, i.e. the hive rds is sync with the table location stored in iceberg s3 metadata. I am not sure how expensive this operation in each table updateOperation. If not much, we should have this validation.  The validate is needed since this is the only place to update & maintain this information, which isn't used by any table operation and won't be caught by other operations. If it's wrong, the data will be deleted by the janitors.

PR #359, opened 2019-08-01
Update QualifiedName to Store the Case-sensitive QualifiedName as well

raveeram:
- Some Data-sources (like Cassandra) have case-sensitive database and table names. Preserve the casing so the corresponding Connector can use the right case-sensitive QualifiedName when needed.

### DIFF
--- a/metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
+++ b/metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
@@ -49,6 +49,9 @@ public final class QualifiedName implements Serializable {
     private Map<String, String> parts;
     private Type type;
 
+    // Null if this QualifiedName was all lower-case to begin with.
+    private QualifiedName caseSensitiveQualifiedName;
+
     private QualifiedName(
         @NonNull final String catalogName,
         @Nullable final String databaseName,
@@ -79,6 +82,10 @@ public final class QualifiedName implements Serializable {
         } else {
             type = Type.CATALOG;
         }
+
+        this.caseSensitiveQualifiedName = createCaseSensitiveQualifiedNameIfNecessary(this.catalogName,
+            databaseName, this.databaseName, tableName, this.tableName, partitionName, this.partitionName,
+            viewName, this.viewName, this.type);
     }
 
     private QualifiedName(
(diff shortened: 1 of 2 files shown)


metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
```diff
@@ -291,6 +298,47 @@ public static QualifiedName ofTable(
         return new QualifiedName(catalogName, databaseName, tableName, null, null);
     }
 
+    private static QualifiedName createCaseSensitiveQualifiedNameIfNecessary(
+        @NonNull final String catalogName,
```
zhljen: I don't have the full context, but I would create a separate function with the non-standardlized qualifiedname as the single input, then it returns the standard qualifiedName as one object. We can also create a new class as the standizedQualified name (which exends the current one) to handle the format and related operations.

raveeram: Agreed, but the expectation is that this is a one-off requirement and I wanted to avoid having to refactor all the QualifiedName constructors in the Controller/Service code to call a new API or a new QualifiedName type. This way, the change is contained inside the QualifiedName itself and the Connector makes a decision on whether to use the case-sensitive version or not.

metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
```diff
@@ -49,6 +49,9 @@
     private Map<String, String> parts;
     private Type type;
 
+    // Null if this QualifiedName was all lower-case to begin with.
+    private QualifiedName caseSensitiveQualifiedName;
```
ajoymajumdar: What is the rationale of not using Optional?

metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
```diff
@@ -291,6 +298,47 @@ public static QualifiedName ofTable(
...
+    private static boolean shouldCreateCaseSensitiveQualifiedName(
+        @NonNull final String databaseName,
+        @NonNull final String standardDatabaseName,
+        @NonNull final String tableName,
+        @NonNull final String standardTableName,
+        @NonNull final String partitionName,
+        @NonNull final String standardPartitionName,
+        @NonNull final String viewName,
+        @NonNull final String standardViewName) {
+        return (!databaseName.isEmpty() && !standardDatabaseName.equals(databaseName))
```
ajoymajumdar: Do we need the isEmpty check?

metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
raveeram: Optional is not serializable and also seems to be a general recommendation that Optional should be used in return values and not for class attributes: https://stackoverflow.com/questions/29033518/is-it-a-good-practice-to-use-optional-as-an-attribute-in-a-class

metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
raveeram: Ah not anymore, was using the NonStandardized string earlier. Removed.

PR #367, opened 2019-09-05
Add a service to provide Aliases for Table Names

raveeram:
- No functionality changes in this PR, just an interface addition. Existing tests should suffice.

### DIFF
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/usermetadata/AliasService.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/usermetadata/AliasService.java
@@ -1,0 +1,39 @@
+/*
+ *  Copyright 2019 Netflix, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+package com.netflix.metacat.common.server.usermetadata;
+
+import com.netflix.metacat.common.QualifiedName;
+import lombok.NonNull;
+
+/**
+ * Alias Service API.
+ *
+ * @author rveeramacheneni
+ * @since 1.3.0
+ */
+public interface AliasService {
+    /**
+     * Returns the original table name if present.
+     *
+     * @param tableAlias the table alias.
+     * @return the original name if present or the alias otherwise.
+     */
+    default QualifiedName getTableName(@NonNull final QualifiedName tableAlias) {
+        return tableAlias;
+    }
+}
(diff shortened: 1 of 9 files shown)


metacat-common-server/src/main/java/com/netflix/metacat/common/server/usermetadata/AliasService.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Alias Service API.
+ *
+ * @author rveeramacheneni
+ * @since 1.3.0
+ */
+public interface AliasService {
+    /**
+     * Returns the original table name if present.
+     *
```
zhljen: I thought the intention is to return the alias name ( if it has one) instead of the original name.

metacat-main/src/main/java/com/netflix/metacat/main/api/RequestWrapper.java
```diff
@@ -87,7 +98,11 @@ public RequestWrapper(@NotNull @NonNull final Registry registry) {
      */
     public QualifiedName qualifyName(final Supplier<QualifiedName> nameSupplier) {
         try {
-            return nameSupplier.get();
+            final QualifiedName name = nameSupplier.get();
+            if (config.isTableAliasEnabled() && name.isTableDefinition()) {
```
zhljen: Why not put the ```name.isTableDefinition()``` check inside the ```aliasService.getTableName()``` API?  The ```getTableName()``` API indicates to me that it's only for tableNames not database, view, or catelog names. It makes sense to me to let the API handle the ```name.isTableDefinition()``` check.  If the API is something like ```getAlias()``` and works for all QualifiedName, check the tableName outside seems a way to control only return the Aliasname for TableName.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/usermetadata/AliasService.java
raveeram: Ah no, it's the opposite (I guess it depends on your frame of reference :)). A request for `hive/default/table` should be mapped to an original name stored in Metacat like `hive/default/table_iceberg` .

metacat-main/src/main/java/com/netflix/metacat/main/api/RequestWrapper.java
ajoymajumdar: Use name.getType() instead of name.isTableDefinition.

ajoymajumdar:
With this change, the response will have the alias name. We need to test this case. If a compute engine calls getTable for table 'abc' and Metacat returns the Table object with name 'xyz', will the engine fail?

metacat-main/src/main/java/com/netflix/metacat/main/api/RequestWrapper.java
raveeram: That makes sense however the `qualifyName(nameSupplier)` API is called for all QualifiedNames, i.e `nameSupplier.get()` could return a Catalog, Database, Table, Partition or View name. Since we only want to check alias for table names, I figured an explicit check in this method would make more sense instead of calling the `aliasService.getTableName()` for every type of QualifiedName.

raveeram: Will do.

raveeram:
@ajoymajumdar that's a good point. For now, I've updated the `getTable()` API to return the `TableDto` with the alias. We'll address other APIs on a case-by-case basis.

PR #372, opened 2019-10-18
On createTable, set the owner(if missing) in serde and definitionMetadata to the user submitting the request.

ajoymajumdar:
If the owner is missing from definitionMetadata only, then set it to that from serde.

### DIFF
--- a/metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
+++ b/metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
@@ -151,15 +151,20 @@ public class TableServiceImpl implements TableService {
     }
 
     private void setOwnerIfNull(final TableDto tableDto, final String user) {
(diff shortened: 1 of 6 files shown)


metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -151,15 +151,20 @@ public TableDto create(final QualifiedName name, final TableDto tableDto) {
...
-        if (!Strings.isNullOrEmpty(user)) {
-            StorageDto serde = tableDto.getSerde();
-            if (serde == null) {
-                serde = new StorageDto();
-                tableDto.setSerde(serde);
-            }
-            if (Strings.isNullOrEmpty(serde.getOwner())) {
-                serde.setOwner(user);
-            }
+        String owner = user;
```
zhljen: Shall we keep this line  `if (!Strings.isNullOrEmpty(user)) {` as the first check?

PR #377, opened 2019-12-02
Changes to support common view

zhljen:


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
```diff
@@ -79,6 +79,15 @@
      * Iceberg table type.
      */
     public static final String ICEBERG_TABLE_TYPE = "ICEBERG";
+    /**
+     * Iceberg table type.
+     */
+    public static final String COMMON_VIEW = "common_view";
```
ajoymajumdar: Is this lower-cased?

zhljen: yes, this is a key in table paramters, e.g.
```
parameters: {
previous_metadata_location: "",
metadata_location: "s3n://netflixxee.metadata.json",
common_view: "true",
table_type: "VIRTUAL_VIEW"
},
```
```

File: metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java [hunk 2]
```diff
@@ -248,16 +258,28 @@ public void updateIcebergTable(final TableInfo tableInfo) {
...
+                return;
+            }
+            if (VIRTUAL_VIEW_TABLE_TYPE.equalsIgnoreCase(tableMetadata.get(PARAM_TABLE_TYPE))
+                && tableMetadata.containsKey(COMMON_VIEW)
+                && Boolean.parseBoolean(tableMetadata.get(DirectSqlTable.COMMON_VIEW))
+            ) {
+                return;
+            }
+        }
+        final String message = String.format("Originally table %s is not type of iceberg ", tableName);
```
ajoymajumdar: Message should have the correct context in terms of iceberg vs common-view.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastTableService.java
```diff
@@ -159,11 +170,36 @@ public void update(final ConnectorRequestContext requestContext, final TableInfo
             } else {
                 directSqlTable.updateIcebergTable(tableInfo);
             }
+        } else if (connectorContext.getConfig().isCommonViewEnabled()
```
ajoymajumdar: We should refactor this to handle common-view separately in a different method (handleCommonView). Also refactor the iceberg code to handleIcebergTable)

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
zhljen: The validation error here is that "we can't update a table that is not an iceberg table or not a common view". I have updated the message.  
I don't think we can't tell if the table should be an iceberg table/common view but it is not so that we can give out the message separately.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/HiveConnectorFastTableService.java
zhljen: Yes, refactored as suggested.

PR #378, opened 2019-12-10
Fixing the common view check condition.

zhljen:


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
```diff
@@ -263,8 +263,7 @@ private void validateTableType(final QualifiedName tableName, final Map<String,
             if (ICEBERG_TABLE_TYPE.equalsIgnoreCase(tableMetadata.get(PARAM_TABLE_TYPE))) {
                 return;
             }
-            if (VIRTUAL_VIEW_TABLE_TYPE.equalsIgnoreCase(tableMetadata.get(PARAM_TABLE_TYPE))
-                && tableMetadata.containsKey(COMMON_VIEW)
+            if (tableMetadata.containsKey(COMMON_VIEW)
```
ajoymajumdar: The two condition check can be replaced with one. Boolean.parseBoolean(tableMetadata.get(COMMON_VIEW))

zhljen: Changed based on comments with a bit code refactor.

metacat-connector-hive/src/test/groovy/com/netflix/metacat/connector/hive/HiveConnectorUtilSpec.groovy
```diff
@@ -42,4 +45,19 @@ class HiveConnectorUtilSpec extends Specification{
         '1'         | 1L
         '010m'      | 600L
     }
+
+    def "Test for check common view" () {
+        given:
+        def tableInfo = TableInfo.builder().metadata(metadata).build()
+        def isView = HiveTableUtil.isCommonView(tableInfo)
+        def isView2 = HiveTableUtil.isCommonView(tableInfo)
```
danielbwatson: I think isView2 can be removed.

metacat-connector-hive/src/test/groovy/com/netflix/metacat/connector/hive/HiveConnectorUtilSpec.groovy
```diff
@@ -42,4 +45,19 @@ class HiveConnectorUtilSpec extends Specification{
...
+        expect:
+        isView == output
+        isView2 == output
+        where:
+        metadata     | output
+        ['common_view' : "true"]    | true
+        ['common_view' : "false"]    | false
+        ['common_view_1' : "false"]    | false
```
danielbwatson: I'd like to add checks for the metadata `[:]`.  Also if metadata is nullable (I haven't checked) we should add the `null` case here as well.

zhljen: Added more cases.

metacat-connector-hive/src/test/groovy/com/netflix/metacat/connector/hive/HiveConnectorUtilSpec.groovy
zhljen: I actually didn't notice this and it's supposed to test a different function. I updated the test and function code. I will merge the code after passing the build.

PR #380, opened 2020-02-10
Fixed issue where iceberg call error results in 5xx error. Invalid request should result in 4xx error.

ajoymajumdar:


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableOpWrapper.java
```diff
@@ -85,11 +84,23 @@ public IcebergTableOpWrapper(final ConnectorContext connectorContext,
...
+                } catch (Exception ignored) {
+                    log.warn("Failed cancelling the task that gets the partitions for an iceberg table.");
+                }
+            }
+            if (e instanceof ExecutionException && e.getCause() != null) {
+                //
+                // On execution exception, throw the inner exception. This is added to throw these as 4xx errors
+                // instead of 5xx.
+                //
+                if (e.getCause() instanceof IllegalArgumentException) {
```
zhljen: Per microbot team, we will return 416 as error code for indicating "the request range cannot be satisfied". Not sure if this change will be the one, would you please chat with them? Also, the change will capture the IllegalArgumentException only,   I haven't found the related iceberg code change yet. In case you have it, please point me to it..

ajoymajumdar: Sure, will let them know. This change with throw 400 error.

PR #381, opened 2020-02-12
Fixed issue where getPartitions API filtering for hive partitions with null values failed.

ajoymajumdar:
Added support for hyphen in partition key names in a partition filter expression


metacat-common-server/src/main/java/com/netflix/metacat/common/server/partition/parser/ASTNULL.java
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright 2016 Netflix, Inc.
+ *    Licensed under the Apache License, Version 2.0 (the "License");
```
zhljen: 2020?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/partition/parser/PartitionParserTreeConstants.java
```diff
@@ -14,49 +14,53 @@
...
+{
+  public int JJTFILTER = 0;
+  public int JJTVOID = 1;
+  public int JJTAND = 2;
+  public int JJTOR = 3;
+  public int JJTNOT = 4;
+  public int JJTBETWEEN = 5;
+  public int JJTIN = 6;
+  public int JJTLIKE = 7;
+  public int JJTNULL = 8;
```
zhljen: Looks this is the one added. Shall we add a comment to make it easier to understand?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/partition/parser/ASTNULL.java
ajoymajumdar: Kept it consistent with what we have in other files.

zhljen:
Looks like the majority of the changes are from the parser, feel it's hard to verify from the diff.  The test definitely gives us confidence. One thing I am not sure if iceberg table would need to make any change to support this or even requires this.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/partition/parser/PartitionParserTreeConstants.java
ajoymajumdar: This file is generated.

ajoymajumdar:
> Looks like the majority of the changes are from the parser, feel it's hard to verify from the diff. The test definitely gives us confidence. One thing I am not sure if iceberg table would need to make any change to support this or even requires this.

My understanding is that we pass the filter expression to iceberg. Iceberg should handle it. If it does not, we need to open a bug for iceberg.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/partition/util/PartitionUtil.java
```diff
@@ -57,9 +58,10 @@ public static void getPartitionKeyValues(final String location, final Map<String
                 // value is null/empty string or any other values that cannot be escaped.
                 //
                 if (values[0].equalsIgnoreCase("null")
-                    || values[1].equalsIgnoreCase("null")
-                    || values[1].equalsIgnoreCase(DEFAULT_PARTITION_NAME)) {
+                    || values[1].equalsIgnoreCase("null")) {
                     log.debug("Found 'null' string in kvp [{}] skipping.", part);
+                } else if (values[1].equalsIgnoreCase(DEFAULT_PARTITION_NAME)) {
```
zhljen: I don't quite understand the logic. Are we saying that if null we will skip but if DEFAULT_PARTITION_NAME we will return null? I guess DEFAULT_PARTITION_NAME is the internal representation of null in hive?

ajoymajumdar: Yes, __HIVE_DEFAULT_PARTITION__ is the internal representation of null in hive.

PR #382, opened 2020-02-19
Add a TABLE_RENAME SNS Message Type

raveeram:
- Today, Rename events are clubbed into the TABLE_UPDATE message type. This will help downstream event consumers (like Microbots) avoid processing irrelevant Update messages.


metacat-main/src/main/java/com/netflix/metacat/main/services/notifications/sns/SNSNotificationServiceImpl.java
```diff
@@ -313,28 +317,39 @@ public void notifyOfTableUpdate(final MetacatUpdateTablePostEvent event) {
             Metrics.CounterSNSNotificationTableUpdate.getMetricName());
     }
 
-    private UpdateTableMessage createUpdateTableMessage(
+    private UpdateOrRenameTableMessageBase createUpdateorRenameTableMessage(
```
zhljen: I feel it's better to have two methods createUpdateTableMessage() and createRenameTableMessage(), each of them will call this createUpdateorRenameTableMessage() with the corresponding message type.

raveeram: Thanks Zhen. Normally I'd prefer the same, but in this case, the difference is just a statement (given the somewhat large method signature). I figured this should suffice as a middle-ground.

PR #386, opened 2020-04-16
Fixed 2 issues.

ajoymajumdar:
1. Fixed issue where getPartitions call with a long filter expression was returning an empty list instead of an error. 2. Fixed issue where the hive metastore persistence factory was not picking up the timeout settings defined in the catalog properties.

### DIFF
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/partition/util/FilterPartition.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/partition/util/FilterPartition.java
@@ -78,7 +78,7 @@ public class FilterPartition {
                 }
             } catch (ParseException | TokenMgrError e) {
                 throw new IllegalArgumentException(String.format("Invalid expression: %s", partitionExpression), e);
-            } catch (StackOverflowError e) {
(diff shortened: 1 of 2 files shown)


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlSavePartition.java
```diff
@@ -457,11 +457,11 @@ public void addUpdateDropPartitions(final QualifiedName tableQName, final Table
         static final String PARTITIONS_SELECT_ALL =
             "SELECT P.PART_ID, P.SD_ID, S.SERDE_ID FROM DBS D JOIN TBLS T ON D.DB_ID=T.DB_ID "
                 + "JOIN PARTITIONS P ON T.TBL_ID=P.TBL_ID JOIN SDS S ON P.SD_ID=S.SD_ID "
-                + "WHERE D.NAME=? and T.TBL_NAME=? limit %d";
+                + "WHERE D.NAME=? and T.TBL_NAME=? limit %d FOR UPDATE";
```
raveeram, line 460: perhaps pull this out into a separate change as discussed.

PR #389, opened 2020-05-02
Added feature to fail rename or delete of a table if it is tagged not to be renamed or deleted

ajoymajumdar:


metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -239,6 +252,20 @@ public TableDto deleteAndReturn(final QualifiedName name, final boolean isMView)
...
 
+    private boolean hasTags(@Nullable final TableDto tableDto, final Set<String> hasTags) {
+        if (!hasTags.isEmpty() && tableDto != null) {
+            final Set<String> tags = getTableTags(tableDto.getDefinitionMetadata());
+            if (!tags.isEmpty()) {
+                for (String t: hasTags) {
+                    if (tags.contains(t)) {
+                        return true;
+                    }
+                }
```
zhljen, line 263: Consider using tags.retainAll(hasTags)?

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -342,6 +369,13 @@ public void rename(
             .includeDefinitionMetadata(true)
             .includeDataMetadata(true)
             .build()).orElseThrow(() -> new TableNotFoundException(oldName));
+
+        // Fail if the table is tagged not to be renamed.
+        if (hasTags(oldTable, config.getNoTableRenameOnTags())) {
+            throw new IllegalArgumentException(
+                String.format("Table %s cannot be renamed because it is tagged.", oldName));
+        }
```
zhljen, line 377: adding the tag that fails this in the message?

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -239,6 +252,20 @@ public TableDto deleteAndReturn(final QualifiedName name, final boolean isMView)
         return tableDto;
     }
 
+    private boolean hasTags(@Nullable final TableDto tableDto, final Set<String> hasTags) {
+        if (!hasTags.isEmpty() && tableDto != null) {
+            final Set<String> tags = getTableTags(tableDto.getDefinitionMetadata());
+            if (!tags.isEmpty()) {
+                for (String t: hasTags) {
```
raveeram, line 259: why not use the intersect()(or retainAll?) equivalent instead

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
ajoymajumdar: Used this since it iterates and exits out even if one matches.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
ajoymajumdar: Used this since it iterates and exits out even if one matches. Using intersect or retailAll is much cleaner than what it is now.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
ajoymajumdar: Will add.

PR #392, opened 2020-06-02
Add support for caching of iceberg tables.

ajoymajumdar:


ajoymajumdar:
> (1) There're two conditions to enable the iceberg table metdata cache, which are
> metacat.cache.enabled, and config.isIcebergCacheEnabled(), I am not sure why the first one is needed. Maybe have a more granular control? Is there any way to detect the setting, e.g. in case we miss the first one.
Yes, the first one is a configuration for the whole application. The second is a more granular one just for iceberg tables. No there is nothing to detect. If the first one is not enabled, then automatically the second one is ignored.
> (2) How the cache eviction behavior? The cache usage here in my view is "get-if-not-found-then- proceed-and-put-eventually, and what will happen for the concurrent threads trying to update/evict the same data. My guess that there might be stale data depending on if the cache copy was the last one. Trying to understand the sequence and impact.
With iceberg, state does not change for specific snapshot referred by the manifest file location. Hence there is not explicit eviction. Cached objects will expire based on the ttl (default is 1 day). For concurrent execution, the cache will get updated multiple times but there will be no stale data.

zhljen:
In concurrent threads case, the cache can get stale data as it doesn't lock anything by default. Maybe we can consider `sync="true"?

ajoymajumdar:
Yes, I am aware of it. It does it for puts and gets. I preferred not to use it since our updates will do a get that will have the cache ready for subsequent requests.

zhljen:
Sure, my understanding when two "updates" happens, the sequence of afterward "get" is uncertain. Maybe I am over-indexed the problem.

PR #411, opened 2020-10-14
Fixed issue where SNS HTTPClient cannot be recovered once shutdown. 

ajoymajumdar:
Creating a new instance of SNSClient when the underlying HTTPClient pool is shutdown.


metacat-main/src/main/java/com/netflix/metacat/main/services/notifications/sns/SNSNotificationServiceImpl.java
```diff
@@ -401,19 +403,44 @@ private void publishNotification(
...
+                .filter(ex -> ex instanceof IllegalStateException
+                    && ex.getMessage().contains("Connection pool shut down"))
+                .findFirst()
+                .map(ex -> {
+                    log.warn("SNS HTTP connection pool is down. It will be restarted.");
+                    try {
+                        this.client.shutdown();
+                    } catch (Exception exception) {
+                        log.warn("Failed shutting down SNS client.", exception);
+                    }
```
raveeram, line 440: Is this going to be thread-safe? Perhaps we need a lock?

ajoymajumdar: Good point. Will need to handle it.

raveeram (approved):

PR #415, opened 2020-11-17
Fix the issue where view creation from iceberg table failed because of field type mismatch.

ajoymajumdar:


metacat-main/src/main/java/com/netflix/metacat/main/services/impl/MViewServiceImpl.java
```diff
@@ -168,6 +184,19 @@ public TableDto createAndSnapshotPartitions(final QualifiedName name,
         return result;
     }
 
+    /**
+     * check if the table is an Iceberg Table.
+     *
+     * @param tableDto table dto
+     * @return true for iceberg table
+     */
+    public static boolean isIcebergTable(final TableDto tableDto) {
```
raveeram, line 193: Looks like this check exists in HiveTableUtil already, should we reuse that?

raveeram (approved):

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/MViewServiceImpl.java
ajoymajumdar: Its in the connector. I do not want to have a direct dependency on the method from this module to a connector module. Also as I commented, this method will be removed as we do not need the condition check.

PR #421, opened 2021-02-26
Add an async post-update Iceberg updateTable event

raveeram:


metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -458,26 +459,33 @@ public TableDto updateAndReturn(final QualifiedName name, final TableDto tableDt
...
-                    .disableOnReadMetadataIntercetor(false)
-                    .includeInfo(true)
-                    .includeDataMetadata(true)
-                    .includeDefinitionMetadata(true)
-                    .build()).orElse(tableDto);
-        } catch (Exception e) {
-            handleException(name, ignoreErrorsAfterUpdate, "getTable", e);
-        }
+        // ignoreErrorsAfterUpdate is currently set only for iceberg tables
+        if (config.isUpdateIcebergTableAsyncPostEventEnabled() && ignoreErrorsAfterUpdate) {
```
ajoymajumdar, line 463: Will this apply to all tables?

raveeram: Ah no, just iceberg tables. ignoreErrorsAfterUpdate was added for a similar purpose and is set to true only for Iceberg. We're re-using that flag here.

ajoymajumdar (approved):

PR #437, opened 2021-05-17
Added API tableExists that checks if a table exists. If the table exists, the response is a HTTP 200. If the table does not exists, the response is a HTTP 404.

ajoymajumdar:


metacat-client/src/main/java/com/netflix/metacat/client/api/MetacatV1.java
```diff
@@ -304,6 +306,39 @@ TableDto getTable(
...
     );
 
+    /**
+     * Returns true, if table exists.
+     * @param catalogName catalog name
+     * @param databaseName database name
+     * @param tableName table name
+     * @return true, if table exists.
+     */
+    default boolean doesTableExists(String catalogName, String databaseName, String tableName) {
```
raveeram, line 316: probably call it doesTableExist for grammatical correctness? :)

raveeram (approved):

raveeram (approved):

PR #441, opened 2021-06-25
Added refreshConnectionPool functionality in DataSourceManager.

ArtfulCoder:
This functionality can be useful when the DataSource needs to be connected to a different underlying data source (eg. a different MySQL Store)

### DIFF
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/util/DataSourceManager.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/util/DataSourceManager.java
@@ -14,6 +14,7 @@
 package com.netflix.metacat.common.server.util;
 
 import com.google.common.collect.Maps;
+import java.sql.SQLException;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.tomcat.jdbc.pool.DataSourceFactory;
 import org.apache.tomcat.jdbc.pool.DataSourceProxy;
@@ -83,6 +84,24 @@ public final class DataSourceManager {
      */
     public DataSource get(final String name) {
         return dataSources.get(name);
+    }
+
+    /**
+     * Refresh the Connection Pool of the data source.
+     * @param name of the DataSource whose ConnectionPool has to be refreshed.
+     * @throws SQLException if pool cannot be re-created.
+     * @return true if success. false otherwise.
+     */
+    public synchronized boolean refreshConnectionPool(final String name) throws SQLException {
+        final DataSource source = dataSources.get(name);
+        if (!(source instanceof DataSourceProxy)) {
+            return false;
+        }
+
+        final DataSourceProxy proxy = (DataSourceProxy) source;
+        proxy.close(true);
+        proxy.createPool();
+        return true;
     }
 
     private synchronized void createDataSource(final String name, final Map props) {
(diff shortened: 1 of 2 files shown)


raveeram (approved):

metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
```diff
@@ -45,10 +45,7 @@
     private final String partitionName;
     private final String tableName;
     private final String viewName;
-    private String qualifiedName;
```
ajoymajumdar, line 48: Why do you not want to store this?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/util/DataSourceManager.java
```diff
@@ -85,6 +86,24 @@ public DataSource get(final String name) {
...
+     * @throws SQLException if pool cannot be re-created.
+     * @return true if success. false otherwise.
+     */
+    public synchronized boolean refreshConnectionPool(final String name) throws SQLException {
+        final DataSource source = dataSources.get(name);
+        if (!(source instanceof DataSourceProxy)) {
+            return false;
+        }
+
+        final DataSourceProxy proxy = (DataSourceProxy) source;
```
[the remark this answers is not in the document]
ArtfulCoder: i think callers can do that based on their needs..

metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
ArtfulCoder: unfortunately, this field is not populated in a thread-safe manner.
concurrency is not given any thought while populating this field.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/util/DataSourceManager.java
```diff
@@ -85,6 +86,24 @@ public DataSource get(final String name) {
         return dataSources.get(name);
     }
 
+    /**
+     * Refresh the Connection Pool of the data source.
+     * @param name of the DataSource whose ConnectionPool has to be refreshed.
+     * @throws SQLException if pool cannot be re-created.
+     * @return true if success. false otherwise.
+     */
+    public synchronized boolean refreshConnectionPool(final String name) throws SQLException {
```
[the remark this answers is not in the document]
ArtfulCoder: I do. I couldn't find existing tests to build on.
lets not make this a commit blocker ?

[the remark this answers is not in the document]
raveeram: Oh yes, for sure. Sorry forgot to resolve the comments.

metacat-common/src/main/java/com/netflix/metacat/common/QualifiedName.java
ajoymajumdar: Agreed. We should have these fields and populate these in the constructor.

ArtfulCoder: is this a known performance bottleneck ?

ajoymajumdar: No. Since this class is immutable, It is better to initialize these in the constructor. What do you think?

ArtfulCoder: If its not a perf bottleneck, I would leave it alone..
otherwise, we would start creating the string rep for all immutable classes..

lets avoid pre-mature optimizations..

ajoymajumdar (approved):

PR #443, opened 2021-07-08
Avoid rewrapping MetacatExceptions in another MetacatException

ArtfulCoder:


metacat-main/src/main/java/com/netflix/metacat/main/api/RequestWrapper.java
```diff
@@ -207,15 +206,12 @@ public QualifiedName qualifyName(final Supplier<QualifiedName> nameSupplier) {
             final String message = String.format("%s.%s -- %s usermetadata operation failed for %s", e.getMessage(),
                 e.getCause() == null ? "" : e.getCause().getMessage(), resourceRequestName, name);
             throw new MetacatUserMetadataException(message);
-        } catch (MetacatUnAuthorizedException e) {
+        } catch (MetacatException e) {
```
raveeram, line 209: Do we want to update the other `processRequest` method with the same behavior?

raveeram (approved):

metacat-main/src/main/java/com/netflix/metacat/main/api/RequestWrapper.java
ArtfulCoder: updated diff to incorporate your feedback.

raveeram (approved):

PR #452, opened 2021-08-20
add Keysets to to table dto / info 

zcxcasd:
adding keysets to table dto + info for jdbc connector tables

sample output from http://localhost:7001/mds/v1/catalog/nrdportal/database/nrdportalprod/table/bf_cert_projects like below:

    "keys":
    {
        "partition":
        [],
        "primary":
        [
            {
                "name": "PRIMARY",
                "fields":
                [
                    "id"
                ]
            }
        ],
        "sort":
        [],
        "index":
        []
    },


metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/model/KeyInfo.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Key Info.
+ */
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+@EqualsAndHashCode(callSuper = false)
+public class KeyInfo extends BaseDto {
```
ajoymajumdar, line 20: Use BaseInfo

ajoymajumdar (approved):

PR #463, opened 2021-10-20
Added polaris connector project.

ArtfulCoder:


metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisDatabaseEntity.java
```diff
@@ -0,0 +1,44 @@
...
+import lombok.ToString;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Version;
+
+/**
+ * Entity class for Database object.
+ */
+@Getter
+@Setter
```
rsyncv, line 23: Why have all fields expose setter? Forcing setting of most required fields at construction time and preferring immutable objects when possible should be the preferred pattern.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnector.java
```diff
@@ -0,0 +1,25 @@
...
+
+import lombok.RequiredArgsConstructor;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.transaction.annotation.Transactional;
+
+/**
+ * This class exposes APIs for CRUD operations.
+ */
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
+public class PolarisConnector {
```
rsyncv, line 11: We might break this into multiple services such as DatabaseService, TableService, and so on. This is pending some design decisions to be made about how much we align with vs break away from the current patterns that each connector follows.

ArtfulCoder: lets leave it as is for now and we can iterate over it.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisDatabaseEntity.java
rsyncv: Using the jpa option of field based instead of property based access makes the setters not needed. Making objects more immutable where possible is a better pattern to follow, for example some fields like generated primary key should purely be managed by the persistence provider and never be set by the application so making the fields private and not exposing any way to set them better ensures against incorrect access.

ArtfulCoder: my question was simply if jpa framework allowed it. 
i tried it and it works. 
PR updated.

rsyncv (dismissed):
Thanks for making the changes looks good

rsyncv (approved):

PR #475, opened 2021-11-01
Added repos and entities packages. Added PolarisStoreService interface.

ArtfulCoder:


metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/store/PolarisConnector.java
```diff
@@ -15,7 +19,7 @@
  * This class exposes APIs for CRUD operations.
  */
 @RequiredArgsConstructor(onConstructor = @__(@Autowired))
-public class PolarisConnector {
+public class PolarisConnector implements PolarisStoreService {
```
raveeram, line 22: Should we rename `PolarisConnector` to something like `PolarisStoreConnector`? Since we already have other connector classes with a similar name.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/store/PolarisStoreService.java
```diff
@@ -0,0 +1,41 @@
...
+ * Interface methods for Polaris Store CRUD access.
+ */
+public interface PolarisStoreService {
+
+    /**
+     * Creates a database entry.
+     * @param databaseName database name
+     * @return Polaris Database entity.
+     */
+    PolarisDatabaseEntity createDatabase(String databaseName);
```
raveeram, line 17: Could this just be the `QualifiedName`?

ArtfulCoder: it can't since it doesnt use catalog name..

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/store/PolarisConnector.java
ArtfulCoder: sure.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/store/PolarisStoreService.java
raveeram: Ah yeah, there are `QualifiedName` APIs to extract out the relevant name parts though. It's the archetype we follow in interfaces in other parts of the codebase and could perhaps be useful from an extensibility standpoint.

ArtfulCoder: from a  method signature that takes QN, it is not obvious that we are not using the catalog.
with this signature, we explicitly call out the parts that we need..

ArtfulCoder: how about we take this offline so that Jeff is unblocked.

raveeram: Makes sense, either way should be fine. Just that this breakdown should happen somewhere in the stack and was thinking it'd be rather useful to do it at the lowest layer. Say tomorrow we might want to use some of the QN methods at this layer too or start logging it (we typically debug by searching for "prodhive/foo/bar") or storing it in a separate layer etc.

raveeram (approved):

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/store/PolarisStoreService.java
```diff
@@ -0,0 +1,68 @@
...
+
+import com.netflix.metacat.connector.polaris.store.entities.PolarisDatabaseEntity;
+import com.netflix.metacat.connector.polaris.store.entities.PolarisTableEntity;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Interface methods for Polaris Store CRUD access.
+ */
+public interface PolarisStoreService {
```
rsyncv, line 11: Could a delete method for database also be added?

ArtfulCoder: sure! will add in next pr. will get it merged and iterate.

PR #477, opened 2021-11-04
Add entity and storage classes for user metadata

raveeram:
- Added basic tests. Will add storage APIs in the next PR.


ArtfulCoder (commented):
superbly done!

metacat-metadata/src/functionalTest/java/com/netflix/metacat/metadata/store/data/repositories/CrdbDataMetadataRepositoryTests.java
```diff
@@ -0,0 +1,18 @@
+//CHECKSTYLE:OFF
```
rsyncv, line 1: What's the rationale for opting the tests out of checkstyle?

metacat-metadata/src/main/java/com/netflix/metacat/metadata/store/data/converters/ObjectNodeConverter.java
```diff
@@ -0,0 +1,40 @@
...
+        this.metacatJson = metacatJson;
+    }
+
+    @Override
+    public String convertToDatabaseColumn(final ObjectNode attribute) {
+        return attribute == null ? null : attribute.toString();
+    }
+
+    @Override
+    public ObjectNode convertToEntityAttribute(final String dbData) {
```
rsyncv, line 37: Would we want error handling in case of invalid json in this converter or just propagate any exception for higher layer to handle?

metacat-metadata/src/main/java/com/netflix/metacat/metadata/store/data/entities/BaseEntity.java
```diff
@@ -0,0 +1,54 @@
...
+import javax.persistence.Version;
+
+/**
+ * Represents a basic metadata entity.
+ *
+ * @author rveeramacheneni
+ */
+@MappedSuperclass
+@Getter
+@Setter
```
rsyncv, line 28: Would it be a better pattern to use the `Setter` tag more sparingly (apply to specific fields instead of the whole class) and prefer immutability when possible, eg for fields like `id` which shouldn't be manually set.

metacat-metadata/src/test/resources/h2db/schema.sql
```diff
@@ -0,0 +1,29 @@
...
+    uri varchar(4000) not null,
+    data json,
+    version bigint not null,
+    is_deleted bit default false not null,
+    created_by varchar(255) not null,
+    created_date timestamp(3) not null,
+    last_updated_by varchar(255) not null,
+    last_updated_date timestamp(3) not null,
+    primary key (id)
+);
```
rsyncv, line 28: Should unique key constraint be added here as well for `uri` and `name` for these h2db versions of the schemas?

metacat-metadata/src/main/java/com/netflix/metacat/metadata/store/data/converters/QualifiedNameConverter.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @author rveeramacheneni
+ */
+@Slf4j
+@Converter(autoApply = true)
+@SuppressWarnings("PMD")
+public class QualifiedNameConverter implements AttributeConverter<QualifiedName, String> {
+
+    @Override
+    public String convertToDatabaseColumn(final QualifiedName attribute) {
```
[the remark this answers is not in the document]
raveeram: Yes, the column values would be `hive/foo/bar` for example

metacat-metadata/src/functionalTest/java/com/netflix/metacat/metadata/store/data/repositories/CrdbDataMetadataRepositoryTests.java
raveeram: Out checkstyle policies already exclude the test package just from an ease of development perspective I guess. we need to update them to exclude the functionalTest package as well.

metacat-metadata/src/main/java/com/netflix/metacat/metadata/store/data/converters/ObjectNodeConverter.java
raveeram: Would perhaps be best to propagate them to the caller so we don't end up storing invalid json in the first place.

metacat-metadata/src/test/resources/h2db/schema.sql
raveeram: It should be, thanks, will update.

metacat-metadata/src/main/java/com/netflix/metacat/metadata/store/data/entities/BaseEntity.java
raveeram: That is true, allowing for setters might help with testing but fields like `id` could perhaps be always immutable. Will update.

metacat-metadata/src/functionalTest/resources/crdb/schema.sql
```diff
@@ -0,0 +1,30 @@
...
+drop table if exists data_metadata;
+
+create table definition_metadata (
+    id UUID NOT NULL DEFAULT gen_random_uuid(),
+    name STRING(1024) not null,
+    data JSON,
+    version INT not null,
+    is_deleted BOOL default FALSE not null,
+    created_by STRING(255) not null,
+    created_date TIMESTAMP not null,
```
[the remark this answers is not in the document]
raveeram: From the CRDB documentation looks like the recommendation is to use `TIMESTAMP` or `TIMESTAMPZ` and in Postgres they're stored as `bigint`, I'm trying to find out how CRDB does it. I'll keep this for now but update if it turns out to be more efficient to use bigint in crdb as well.

rsyncv (approved):

rsyncv (approved):

PR #485, opened 2021-11-20
Update version column when metadataLocation is updated.

ArtfulCoder:


raveeram (approved):
Thanks abhi, these changes make sense to me!

PR #491, opened 2021-12-15
Polaris table update api with tests

rsyncv:


metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
```diff
@@ -182,16 +188,57 @@ public TableInfo get(final ConnectorRequestContext requestContext, final Qualifi
...
+            final String newLoc = newTableMetadata.get(DirectSqlTable.PARAM_METADATA_LOCATION);
+            if (StringUtils.isBlank(prevLoc) || StringUtils.isBlank(newLoc)) {
+                final String message = String.format(
+                    "Invalid metadata for %s. Provided previous %s or new %s location is empty.",
+                    name, prevLoc, newLoc);
+                log.error(message);
+                throw new InvalidMetaException(name, message, null);
+            }
+            // optimistically attempt to update metadata location
+            final boolean updated = polarisConnector.updateTableMetadataLocation(
```
ajoymajumdar, line 209: Are we not storing other table properties?

ajoymajumdar, line 209: What happens if there is a timeout?

rsyncv: Not currently, we might start storing previous location as a separate field though.

rsyncv: We are planning to add a retry layer within the store layer, so this service layer does not need to handle those.

rsyncv: We'll have a sync to see if some of the other current use cases can be moved to use the Iceberg map instead.

raveeram (approved):

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
```diff
@@ -182,16 +188,57 @@ public TableInfo get(final ConnectorRequestContext requestContext, final Qualifi
...
+                name.getDatabaseName(), name.getTableName(), prevLoc, newLoc);
+            // if succeeded then done, else try to figure out why and throw corresponding exception
+            if (updated) {
```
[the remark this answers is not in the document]
rsyncv: will do, making the change on the subsequent commit

PR #494, opened 2021-12-29
Add table CRUD smoke tests for Polaris

raveeram:
- Additionally, a few minor fixes for issues found during testing.


ajoymajumdar:
What were the issues fixed?

ajoymajumdar (approved):

rsyncv (approved):
Discussed tradeoffs of this approach offline, good to go

raveeram:
> What were the issues fixed?
They were mainly around ensuring parity with the existing behavior when storing iceberg metadata in HMS like:
- If the proposed metadata location as the existing one, we no longer throw a `PreconditionFailedException`. 
- If the existing metadata location is no longer valid, prevent the table commits from moving forward. 
- Ensure the exception types thrown (InvalidMetaException vs ConnectorException) match what we had before in HMS.

PR #495, opened 2022-01-08
Store the previous_metadata_location for iceberg tables in polaris

raveeram:


metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/store/repos/PolarisTableRepository.java
```diff
@@ -87,8 +87,9 @@ boolean existsByDbNameAndTblName(
      * @return number of rows that are updated.
      */
     @Modifying(flushAutomatically = true, clearAutomatically = true)
-    @Query("UPDATE PolarisTableEntity t SET t.metadataLocation = :newLocation, t.version = t.version + 1 "
-        + "WHERE t.metadataLocation = :expectedLocation AND t.dbName = :dbName AND t.tblName = :tableName")
+    @Query("UPDATE PolarisTableEntity t SET t.metadataLocation = :newLocation, "
+            + "t.previousMetadataLocation = t.metadataLocation, t.version = t.version + 1 "
+            + "WHERE t.metadataLocation = :expectedLocation AND t.dbName = :dbName AND t.tblName = :tableName")
```
rsyncv, line 92: Recall we were discussing idempotency in the case where `previousMetadataLocation = newLocation`, whether we want it to succeed as a no op or fail. Now that `previousMetadataLocation` field is added, it is possible to add the clause `metadataLocation != newLocation` in the filter conditions if the latter behavior is more desirable.

metacat-connector-polaris/src/functionalTest/resources/schema.sql
```diff
@@ -12,7 +12,8 @@ create table TBLS (
   id uuid default gen_random_uuid() not null primary key,
   db_name varchar(255) not null,
   tbl_name varchar(255) not null,
-  metadata_location varchar(1024),
+  previous_metadata_location varchar(8192),
```
rsyncv, line 15: Is `8192` the more correct upper bound size? If so we have to change it in the test crdb as well.

rsyncv (approved):

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/store/repos/PolarisTableRepository.java
raveeram: Discussed this offline, for now, we'll keep the existing logic of having a "retry" essentially be a no-op.

metacat-connector-polaris/src/functionalTest/resources/schema.sql
raveeram: Ah yeah, it's just a better guess. Thanks, updated it in test crdb as well.

rsyncv (approved):

rsyncv (approved):

PR #499, opened 2022-01-15
Parse table uri to include in serde and add serde tests

rsyncv:


metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/mappers/PolarisTableMapper.java
```diff
@@ -36,6 +36,7 @@ public PolarisTableMapper(final String catalogName) {
      */
     @Override
     public TableInfo toInfo(final PolarisTableEntity entity) {
+        final int uriIndex = entity.getMetadataLocation().indexOf("/metadata/");
```
raveeram, line 39: Do we want to make this a const string?

raveeram (approved):

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/mappers/PolarisTableMapper.java
rsyncv: Ok

raveeram (approved):

PR #508, opened 2022-05-02
Add flag to only fetch metadata location for iceberg tables

raveeram:


metacat-common-server/src/main/java/com/netflix/metacat/common/server/api/v1/MetacatV1.java
```diff
@@ -71,14 +71,41 @@ default TableDto getTable(
...
+     * Get the table.
+     *
+     * @param catalogName                 catalog name
+     * @param databaseName                database name
+     * @param tableName                   table name.
+     * @param includeInfo                 true if the details need to be included
+     * @param includeDefinitionMetadata   true if the definition metadata to be included
+     * @param includeDataMetadata         true if the data metadata to be included
+     * @param includeInfoDetails          true if the more info details to be included
+     * @param includeMetadataLocationOnly true if only metadata location needs to be included
```
ajoymajumdar, line 97: If true, the other flags are ignored?

ajoymajumdar, line 97: If true, the other flags are ignored?

ajoymajumdar:
This change will impact the caching strategy. Does it handle the case where a request  with metadataLocationOnly ends up caching TableDto with no schema info and the subsequent request to fetch all will end up not getting all the information about the table?

ajoymajumdar:
This change will impact the caching strategy. Does it handle the case where a request  with metadataLocationOnly ends up caching TableDto with no schema info and the subsequent request to fetch all will end up not getting all the information about the table?

metacat-common-server/src/main/java/com/netflix/metacat/common/server/api/v1/MetacatV1.java
raveeram: Thanks, updated.

raveeram:
> 

Great point, updated the caching strategy to handle this scenario and also to not needlessly evict any existing keys. Tested manually now but we'll look into adding an integration test for this logic later.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/ConnectorTableServiceProxy.java
```diff
@@ -138,7 +148,10 @@ public void rename(
      * @param tableInfo table object
      * @return true if errors after this should be ignored.
      */
-    @CacheEvict(key = "'table.' + #name")
+    @Caching(evict = {
+            @CacheEvict(key = "'table.' + #oldName"),
```
ajoymajumdar, line 152: Incorrect key.

raveeram: thanks, fixed.

ajoymajumdar (approved):

PR #509, opened 2022-05-13
Implement get table names api for polaris connector

rsyncv:


metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
```diff
@@ -357,4 +358,32 @@ public TableInfo getIcebergTable(final QualifiedName tableName,
             this.icebergTableHandler.getIcebergTable(tableName, tableMetadataLocation, includeInfoDetails);
         return connectorConverter.fromIcebergTableToTableInfo(tableName, icebergTable, tableMetadataLocation, info);
     }
+
+    @Override
+    public List<QualifiedName> getTableNames(
+        final ConnectorRequestContext context,
+        final QualifiedName name,
+        final String filter,
```
ajoymajumdar, line 366: `filter` is ignored? Probably throw a NotSupported error if filter is not empty.

rsyncv: Yeah we decided not to support this general HMS filter expression for Polaris. Good suggestion, will update to throw exception.

raveeram (dismissed):

rsyncv:
Ok, updated to log the filter.

ajoymajumdar (approved):

PR #519, opened 2022-10-19
remove dependency on servo

brharrington:
Servo has been deprecated for a long time and isn't really needed for this project. This change removes the dependency.


ajoymajumdar (approved):
Thanks for the change.

PR #520, opened 2022-10-26
Add a getConnection API to HiveConnector fast table service

swaranga-netflix:
Adding an APi to get the underlying JDBC connection from the Hive fast table service.

### DIFF
--- a/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
+++ b/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
@@ -47,6 +47,8 @@ import org.springframework.jdbc.core.SqlParameterValue;
 import org.springframework.transaction.annotation.Propagation;
 import org.springframework.transaction.annotation.Transactional;
 
+import java.sql.Connection;
+import java.sql.SQLException;
 import java.sql.Types;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/sql/DirectSqlTable.java
```diff
@@ -137,6 +140,18 @@ public DirectSqlTable(
         this.config = connectorContext.getConfig();
     }
 
+    /**
+     * Returns the Jdbc connection of the underlying database.
+     *
+     * @return the Jdbc connection of the underlying database
+     * @throws SQLException if the connection could not be fetched
+     * @throws NullPointerException if no data source has been configured
```
rsyncv, line 148: Knowingly throwing null pointer exception is an antipattern, can this instead do a null check and throw a more informative exception

swaranga-netflix: I debated between a few options 
- an NPE with a descriptive message
- an IllegalArgException: This felt out of place since the callers isn't passing the arguments
- An IllegalStateException: This also seemed less appropriate for a bean configuration error
- A new custom exception class: Seemed overkill for this case as the method is not going to be part of the connector public API.
- An SqlException: Did not fit the case as it is not really database related error

In the end I settled for the NPE with a descriptive error as the change itself is temporary and the message has enough information that the system owner can go and fix the issue.

rsyncv (approved):

PR #524, opened 2023-01-23
Move invalid owner diagnostic logging to table service and include root user as invalid

swaranga-netflix:
Moved the diagnostic logging to TableServiceImpl after all processing of the owner (from the context + serdes) are done.

Also, now including the root user as an invalid user since that seems to be the primary invalid owners we see

### DIFF
--- a/metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
+++ b/metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
@@ -59,6 +59,7 @@ import com.netflix.metacat.main.services.TableService;
 import com.netflix.spectator.api.Registry;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang.StringUtils;
 import org.springframework.web.context.request.RequestAttributes;
 import org.springframework.web.context.request.RequestContextHolder;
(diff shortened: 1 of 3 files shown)


metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -172,6 +180,66 @@ public TableDto create(final QualifiedName name, final TableDto tableDto) {
...
+     */
+    private void logOwnershipDiagnosticDetails(final QualifiedName name, final TableDto tableDto) {
+        try {
+            final String tableOwner = Optional.ofNullable(tableDto.getDefinitionMetadata())
+                                          .map(node -> node.get("owner"))
+                                          .map(owner -> owner.get("userId"))
+                                          .map(JsonNode::textValue)
+                                          .orElse(null);
+
+            if (tableOwner == null || "metacat".equals(tableOwner) || "root".equals(tableOwner)) {
```
rsyncv, line 197: Should this also check for empty string

rsyncv (dismissed):

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
swaranga-netflix: Although unlikely probably safer to add. Added.

rsyncv (approved):

PR #525, opened 2023-01-24
Fix owner selection logic in table create

swaranga-netflix:


metacat-common/src/main/java/com/netflix/metacat/common/dto/TableDto.java
```diff
@@ -89,6 +91,14 @@ public QualifiedName getDefinitionName() {
         return name;
     }
 
+    @JsonIgnore
+    public Optional<String> getTableOwner() {
+        return Optional.ofNullable(definitionMetadata)
+                   .map(definitionMetadataJson -> definitionMetadataJson.get("owner"))
+                   .map(ownerJson -> ownerJson.get("userId"))
+                   .map(JsonNode::textValue);
```
rsyncv, line 99: Is the `orElse(null)` no longer needed here

ajoymajumdar (approved):

metacat-common/src/main/java/com/netflix/metacat/common/dto/TableDto.java
swaranga-netflix: The method returns an `Optional` to signal to the callers that this can be null. Hence the orElse is only needed at the call-site.

PR #527, opened 2023-01-30
Fix NPE in RemoveTags API

raveeram:


metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlTagService.java
```diff
@@ -278,6 +278,9 @@ public void remove(final QualifiedName name, final Set<String> tags, final boole
                 params.toArray());
             if (updateUserMetadata) {
                 final TagItem tagItem = get(name);
+                if (tagItem == null || tagItem.getValues().isEmpty()) {
```
ajoymajumdar, line 281: Looks like after the delete statement, the get will always return null? The code should do a get before the delete statement.

raveeram: Yup thanks, perhaps not the get, but the getValues() could be empty always. Fixed.

ajoymajumdar (approved):

rsyncv (approved):

PR #532, opened 2023-02-14
Implement partition APIs for polaris connector

swaranga-netflix:
* Implemented the read-only APIs except the getPartitionNames.
* All write APIs throw unsupported operation exception
* Moved some common Iceberg specific code to IcebergTableHandler from the hive connector so we can reuse them in the polaris connector
* Split the checkstyle rules for test with some rules relaxed:
  * Allow static imports
  * No need of Javadoc for tests
  * Relax rules on final local variable and method parameters

### DIFF
--- a/codequality/checkstyle/checkstyle-test.xml
+++ b/codequality/checkstyle/checkstyle-test.xml
@@ -50,11 +50,6 @@
         <!-- Check that finds import statements that use the * notation. -->
         <module name="AvoidStarImport"/>
 
-        <!-- Check that finds static imports. -->
-        <module name="AvoidStaticImport">
-            <property name="excludes" value="org.mockito.Mockito.*,org.assertj.core.api.Assertions.*"/>
-        </module>
-
         <!-- Checks that constant names conform to a format specified by the format property. -->
         <module name="ConstantName"/>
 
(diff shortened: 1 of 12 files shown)


codequality/checkstyle/checkstyle-test.xml
```diff
@@ -0,0 +1,261 @@
+<?xml version="1.0"?>
```
rsyncv, line 1: Would have been nice if this could have been refactored as a base config and override config to avoid duplication, but seems there is no standard paved path way currently.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorPartitionService.java
```diff
@@ -0,0 +1,120 @@
...
+        final List<PartitionInfo> partitions = icebergTableHandler.getPartitions(
+            tableInfo,
+            context,
+            null,
+            Collections.singletonList(partitionName.getPartitionName()),
+            null
+        );
+
+        return partitions.stream()
+                   .filter(partitionInfo -> partitionInfo.getName().equals(partitionName))
```
rsyncv, line 116: Seems it should be safe without this filter since the name filter was already passed in the earlier call, though checking again here is also fine.

rsyncv (approved):

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorPartitionService.java
swaranga-netflix: Good eye; does not harm except for the extra allocations and checks. Will remove it next time I make a change in this repo.

codequality/checkstyle/checkstyle-test.xml
swaranga-netflix: Agreed, xml import does not work with Checkstyle documents since both need to be full XML documents

PR #533, opened 2023-02-17
Add user id validation for create and update API

swaranga-netflix:
Introduce an interface for user validation and add a default implementation that checks for know invalid users. In metacat-netflix, we will override it with a dynamic check from a remote source.

We will also remove this check from the event listener and will now do it in the table service so that we can guarantee access to all the thread locals like Http headers and the request context.

We are now also adding this check in the update API if the request contains a change to the owner field

### DIFF
--- a/metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
+++ b/metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
@@ -56,20 +56,14 @@ import com.netflix.metacat.main.manager.ConnectorManager;
 import com.netflix.metacat.main.services.DatabaseService;
 import com.netflix.metacat.main.services.GetTableNamesServiceParameters;
 import com.netflix.metacat.main.services.GetTableServiceParameters;
+import com.netflix.metacat.main.services.OwnerValidationService;
 import com.netflix.metacat.main.services.TableService;
 import com.netflix.spectator.api.Registry;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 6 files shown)


metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -548,6 +472,12 @@ public TableDto updateAndReturn(final QualifiedName name, final TableDto tableDt
             ignoreErrorsAfterUpdate = connectorTableServiceProxy.update(name, converterUtil.fromTableDto(tableDto));
         }
 
+        // we do ownership validation and enforcement only if table owner is set in the dto
+        // because if it is null, we do not update the owner in the existing metadata record
+        if (tableDto.getTableOwner().isPresent()) {
+            ownerValidationService.enforceOwnerValidation("updateTable", name, tableDto);
```
rsyncv, line 478: This `TableServiceImpl` has logic applied commonly across catalogs. To start we are focusing on ownership for hive/iceberg tables, what is your plan to be able to selectively enable validation on certain catalogs?

swaranga-netflix: The enforce method takes in the Qualified name of the table. In the pandora implementation, I was planning to add a check for the two hive catalogs. In future, if we want to do this for all catalogs, we can remove that check.

swaranga-netflix: Having said that, the existing event listener implementation (when it was enabled) did not check for catalog types before checking for the user validity. I would say eventually we want to have it for all catalogs though and the number of calls for the other catalogs should be tiny compared to hive.

rsyncv (approved):

PR #538, opened 2023-03-01
Fix bug with hive column name with space replacement

swaranga-netflix:
We were using the wrong method from the matcher. regionStart actually gives you the start index of the input being matched. The right method to use here is `start` and `end`.

Also added test cases for all the column types for the table that was affected.


metacat-connector-hive/src/test/groovy/com/netflix/metacat/connector/hive/converters/HiveTypeConverterSpec.groovy
```diff
@@ -105,6 +105,20 @@ class HiveTypeConverterSpec extends Specification {
             "struct<field1:string,field2:decimal (38,9 ),field3:bigint>",
             "struct<field1:string,field2:decimal (38),field3:bigint>",
             "struct<field1:string,field2:decimal(38 ),field3:bigint>",
+
+            "struct<prediction_date:int,lower_confidence_amt:decimal(30,2),upper_confidence_amt:decimal(30,2),model_short_name:string>",
```
raveeram, line 109: Nice work with the thorough test cases!

raveeram (approved):

rsyncv (approved):

PR #541, opened 2023-03-07
Metacat throttling refactor

swaranga-netflix:
* Refactor connector manager to allow decorating connector services with common functionality
* Perform throttling at the connector layer
* Split request validation and rate limiting so that only request validation happens in the controller layer

Older related PR: https://github.com/Netflix/metacat/pull/531 which will now be discarded

### DIFF
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorFactoryDecorator.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorFactoryDecorator.java
@@ -23,8 +23,8 @@ public class ConnectorFactoryDecorator implements ConnectorFactory {
         ConnectorCatalogService service = delegate.getCatalogService();
 
         if (isRateLimiterEnabled()) {
-            log.info("Creating rate-limited connector catalog services for connector-type: {}, " +
-                         "plugin-type: {}, catalog: {}, shard: {}",
+            log.info("Creating rate-limited connector catalog services for connector-type: {}, "
+                         + "plugin-type: {}, catalog: {}, shard: {}",
                 connectorContext.getConnectorType(), connectorPlugin.getType(),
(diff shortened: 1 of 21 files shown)


metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorFactoryDecorator.java
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        return service;
+    }
+
+    @Override
+    public ConnectorTableService getTableService() {
+        ConnectorTableService service = delegate.getTableService();
+
+        if (isRateLimiterEnabled()) {
```
rsyncv, line 55: The throttling should be able to be enabled dynamically with FP. Could this be reworked to allow that, for example if service starts up with throttling disabled it should be possible to gradually reenable dynamically per cluster and vice versa.

swaranga-netflix: This throttle config is only to prevent the migration connector from being wrapped with the throttling decorator - this prevents double throttling first at the migration connector and then at the underlying nested connector. The MetacatRateLimiter implementation of the RateLimiter that we will have in the internal repo will be able to handle dynamic updates to throttling config changes.

Only the migrationconnector.properties will have this exempt set to true. So this config is fine as static.

rsyncv: Ok makes sense

rsyncv (commented):
The aim of this change is to move rate limiting after caching, so previously we had discussed between possible choices of moving the throttling to the unified `ServiceProxy` layer vs the connector layer where it selectively wraps individual connector beans. Could you give an explanation for what are the reasons you eventually chose the latter design over the former, for code reviewers as well as for future reference.

rsyncv (approved):

PR #542, opened 2023-03-13
Upgrade to Spring boot 2.7.6

swaranga-netflix:


metacat-main/build.gradle
```diff
@@ -76,9 +77,10 @@ dependencies {
     runtime(project(":metacat-connector-druid"))
 
     runtime("io.springfox:springfox-swagger-ui")
-    runtime("org.webjars:hal-browser")
+    runtime("org.webjars:hal-browser:3325375")
```
rsyncv, line 80: Curious what the reason for this pinning is

metacat-functional-tests/src/functionalTest/groovy/com/netflix/metacat/MetacatSmokeSpec.groovy
```diff
@@ -123,6 +123,16 @@ class MetacatSmokeSpec extends Specification {
             if (metadata != null) {
                 newTable.getMetadata() == null? newTable.setMetadata(metadata): newTable.getMetadata().putAll(metadata)
             }
+
+            if (tableName == 's3-mysql-db') {
```
rsyncv, line 127: Would it make sense to just set the owner for all tables created

metacat-main/build.gradle
swaranga-netflix: The new Netflix BOm does not have recommendations for this artifact, so we have to specify a specific version

metacat-functional-tests/src/functionalTest/groovy/com/netflix/metacat/MetacatSmokeSpec.groovy
swaranga-netflix: It is mandated only for the S3 connector, but let me try and see what breaks.

rsyncv (approved):

PR #543, opened 2023-03-15
Refactor connector decorator

swaranga-netflix:
* In the current implementation, every dependency of the decorators would need to be plumbed into multiple places.
  * For instance, the rate limiter, had to be plumbed through the connector manager and would need similar plumbing in the migration connector factory.
 * With the new implementation, deciding what wrappers to use and what dependencies are needed is confined in the decorating connector factory.

### DIFF
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorFactoryDecorator.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorFactoryDecorator.java
@@ -1,7 +1,5 @@
 package com.netflix.metacat.common.server.connectors;
 
-import com.google.common.base.Supplier;
-import com.google.common.base.Suppliers;
 import com.netflix.metacat.common.server.api.ratelimiter.RateLimiter;
 import lombok.Getter;
 import lombok.NonNull;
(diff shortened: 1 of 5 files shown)


metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorFactoryDecorator.java
```diff
@@ -1,81 +1,86 @@
...
+                connectorContext.getCatalogName(), connectorContext.getCatalogShardName());
 
-        if (isRateLimiterEnabled()) {
-            log.info("Creating rate-limited connector catalog services for connector-type: {}, "
+            final RateLimiter rateLimiter = connectorContext.getApplicationContext().getBean(RateLimiter.class);
+
+            // not all connectors implement these methods;
+            // calling these eagerly may throw an unsupported operation exception;
+            // hence we only memoize these for now and call them at runtime when they are needed
+            catalogService = memoize(
```
rsyncv, line 45: Is the memoize mainly for delaying the eager call, if so could it cause issues due to returning cached result instead of new object each time

swaranga-netflix: Yeah, the memoization was to delay the eager call and should not cause issues because these services are singleton; but after sleeping on it I am thinking it is better not to memoize it and introduce that implicit coupling. Will move back to calling the delegate for every call and wrapping it with the throttling connector.

PR #544, opened 2023-03-19
Delete the storage table when a materialized common view is deleted

raveeram:
- Validated with an e2e manual test, will try to add unit tests.

### DIFF
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/properties/Config.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/properties/Config.java
@@ -459,6 +459,14 @@ public interface Config {
     boolean isCommonViewEnabled();
 
     /**
+     * Whether the underlying storage table should be deleted
+     * for a materialized view.
+     *
+     * @return true if the storage table should be deleted when the common view is.
+     */
+    boolean deleteCommonViewStorageTable();
+
+    /**
      * get Iceberg table refresh metadata location retry number.
      *
      * @return Refresh metadata location retry number.
--- a/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/util/HiveTableUtil.java
+++ b/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/util/HiveTableUtil.java
@@ -22,6 +22,7 @@ import com.netflix.metacat.common.QualifiedName;
 import com.netflix.metacat.common.server.connectors.exception.InvalidMetaException;
 import com.netflix.metacat.common.server.connectors.model.TableInfo;
 import com.netflix.metacat.common.server.util.MetacatContextManager;
+import com.netflix.metacat.common.server.util.MetacatUtils;
 import com.netflix.metacat.connector.hive.sql.DirectSqlTable;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang.StringUtils;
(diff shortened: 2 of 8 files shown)


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/util/HiveTableUtil.java
```diff
@@ -158,27 +159,15 @@ public static TableIdentifier qualifiedNameToTableIdentifier(final QualifiedName
...
-
-    /**
-     * check if the table is a common view.
-     *
-     * @param tableMetadata table metadata map
-     * @return true for common view
-     */
-    public static boolean isCommonView(final Map<String, String> tableMetadata) {
-        return tableMetadata != null && Boolean.parseBoolean(tableMetadata.get(DirectSqlTable.COMMON_VIEW));
+        return tableInfo.getMetadata() != null && MetacatUtils.isCommonView(tableInfo.getMetadata());
```
zhljen, line 168: It looks to me that this should be 
`tableInfo != null && && MetacatUtils.isCommonView(tableInfo.getMetadata())`;  since the parameter is null will be handled inside the isCommonView function.

zhljen (commented):
Looks good to me. I suggest adding a function test for the feature coverage.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -736,6 +751,19 @@ private TableInfo getFromTableServiceProxy(final QualifiedName name,
                 : connectorTableServiceProxy.get(name, getTableServiceParameters, useCache);
     }
 
+    private void deleteCommonViewStorageTable(final QualifiedName viewName,
+                                              final QualifiedName storageTableName) {
+
+        try {
+            Log.warn("Deleting storage table: {} belonging to common view: {}",
+                    storageTableName, viewName);
+            connectorTableServiceProxy.delete(storageTableName);
```
ajoymajumdar, line 760: This should be calling method `TableServiceImpl.deleteAndReturn`

raveeram: Makes sense, updated.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/properties/Config.java
```diff
@@ -458,6 +458,14 @@ public interface Config {
      */
     boolean isCommonViewEnabled();
 
+    /**
+     * Whether the underlying storage table should be deleted
+     * for a materialized view.
+     *
```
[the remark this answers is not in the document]
raveeram: > Looks good to me. I suggest adding a function test for the feature coverage.

Thanks Zhen, added a functional test

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -744,6 +753,19 @@ private TableInfo getFromTableServiceProxy(final QualifiedName name,
                         : connectorTableServiceProxy.get(name, getTableServiceParameters, useCache));
     }
 
+    private void deleteCommonViewStorageTable(final QualifiedName viewName,
+                                              final QualifiedName storageTableName) {
+
+        try {
+            log.warn("Deleting storage table: {} belonging to common view: {}",
+                    storageTableName, viewName);
+            deleteAndReturn(storageTableName, false);
```
zhljen, line 762: This is a recursive call from the view deletion, and will return since the flag=false

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -267,6 +267,20 @@ public TableDto deleteAndReturn(final QualifiedName name, final boolean isMView)
...
+            // If this is a common view, the storage_table if present
+            // should also be deleted.
+            if (MetacatUtils.isCommonView(tableDto.getMetadata())
+                    && config.deleteCommonViewStorageTable()) {
+                final Optional<String> storageTableName = MetacatUtils
+                        .getCommonViewStorageTable(tableDto.getMetadata());
+                if (storageTableName.isPresent()) {
+                    final QualifiedName qualifiedStorageTableName = QualifiedName.ofTable(name.getCatalogName(),
+                            name.getDatabaseName(), storageTableName.get());
+                    deleteCommonViewStorageTable(name, qualifiedStorageTableName);
```
zhljen, line 280: Construct the `qualifiedStorageTableName` using view name as db which is default.

zhljen (approved):

PR #554, opened 2023-05-09
Differentiate log levels between user and system exceptions

rsyncv:


metacat-main/src/main/java/com/netflix/metacat/main/api/ExceptionMapper.java
```diff
@@ -66,18 +67,26 @@ public void handleMetacatException(
...
+            logErrorLevel = true;
             status = HttpStatus.NOT_IMPLEMENTED.value();
         } else if (e instanceof MetacatUserMetadataException) {
             // TODO: This makes no sense
             status = HttpStatus.SEE_OTHER.value();
         } else if (e instanceof MetacatTooManyRequestsException) {
+            logErrorLevel = true;
             status = HttpStatus.TOO_MANY_REQUESTS.value();
         } else if (e instanceof MetacatUnAuthorizedException) {
+            logErrorLevel = true;
```
swaranga-netflix, line 79: Should this be a user exception?

metacat-main/src/main/java/com/netflix/metacat/main/api/ExceptionMapper.java
```diff
@@ -66,18 +67,26 @@ public void handleMetacatException(
...
             status = HttpStatus.FORBIDDEN.value();
         } else {
+            logErrorLevel = true;
             status = HttpStatus.INTERNAL_SERVER_ERROR.value();
         }
-        log.error(e.getLocalizedMessage(), e);
+        if (logErrorLevel) {
+            log.error(e.getLocalizedMessage(), e);
+        } else {
+            log.warn(e.getLocalizedMessage(), e);
```
swaranga-netflix, line 88: We can also probably omit the stack trace for this one.

metacat-main/src/main/java/com/netflix/metacat/main/api/ExceptionMapper.java
```diff
@@ -66,18 +67,26 @@ public void handleMetacatException(
         } else if (e instanceof MetacatNotFoundException) {
             status = HttpStatus.NOT_FOUND.value();
         } else if (e instanceof MetacatNotSupportedException) {
+            logErrorLevel = true;
             status = HttpStatus.NOT_IMPLEMENTED.value();
         } else if (e instanceof MetacatUserMetadataException) {
             // TODO: This makes no sense
             status = HttpStatus.SEE_OTHER.value();
         } else if (e instanceof MetacatTooManyRequestsException) {
+            logErrorLevel = true;
```
swaranga-netflix, line 76: Also probably a user exception?

metacat-main/src/main/java/com/netflix/metacat/main/api/ExceptionMapper.java
rsyncv: Would like to keep logging stack trace for now, this PR is mainly focused on separating the two types so as to be more easily searchable and distinguishable in something like radar.

metacat-main/src/main/java/com/netflix/metacat/main/api/ExceptionMapper.java
rsyncv: Ok.

swaranga-netflix (approved):

PR #555, opened 2023-05-11
Add metrics for get table parameters

swaranga-netflix:


metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -318,6 +318,9 @@ private boolean isEnabledForTableDefinitionMetadataDelete(final QualifiedName ta
     @Override
     public Optional<TableDto> get(final QualifiedName name, final GetTableServiceParameters getTableServiceParameters) {
         validate(name);
+
+        logGetTableMetrics(name, getTableServiceParameters);
```
ajoymajumdar, line 322: We log request metrics in https://github.com/Netflix/metacat/blob/master/metacat-main/src/main/java/com/netflix/metacat/main/api/RequestWrapper.java. Should we add this logic there?

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -509,6 +512,26 @@ public TableDto updateAndReturn(final QualifiedName name, final TableDto tableDt
...
+     *
+     * @param name the qualified name of the table
+     * @param params the get-table parameters
+     */
+    private void logGetTableMetrics(final QualifiedName name,
+                                    final GetTableServiceParameters params) {
+        registry.counter(
+            "metacat.tableService.getTable",
+            "catalog", name.getCatalogName(),
+            "database", name.getDatabaseName(),
```
ajoymajumdar, line 526: From requestWrapper, the registry metric `metacat.server.counter.request` have tags - catalogName, databaseName, tableName.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
swaranga-netflix: Much better; thanks.

ajoymajumdar (approved):

PR #558, opened 2023-05-27
Fix initialization or service controllers

swaranga-netflix:
Tested this locally and verified that until the service starts up, we refuse the connections.

```
2023-05-27T00:47:43.884155564Z: Connection refused
2023-05-27T00:47:44.884738722Z: Connection refused
2023-05-27T00:47:45.885341280Z: Connection refused
2023-05-27T00:47:46.885831217Z: Connection refused
2023-05-27T00:47:47.886312643Z: Connection refused
2023-05-27T00:47:58.076105262Z: Response code: 200
2023-05-27T00:48:05.488582010Z: Response code: 200
2023-05-27T00:48:12.596388501Z: Response code: 200
```

The previous implementation was initializing the catalog only after the context refresh event by when the controllers had already been initialized and started service requests which would lead to 404s. With this change we should not get the 404s.


rsyncv (dismissed):

metacat-main/src/main/java/com/netflix/metacat/main/configs/ServicesConfig.java
```diff
@@ -400,22 +401,25 @@ public MetadataService metadataService(
      * @param connectorManager     Connector manager to use
      * @param threadServiceManager Thread service manager to use
      * @param metacatThriftService Thrift service to use
+     * @param applicationContext the application context
+     *
      * @return The initialization service bean
      */
-    @Bean
+    @Bean(initMethod = "start", destroyMethod = "stop")
```
ajoymajumdar, line 408: This makes sure that the application context is ready before loading the catalogs. How do we make sure that this is initialized since none of the classes have a reference to it.

swaranga-netflix: Spring beans are eagerly initialized. So the bean itself should be initialized. As for the ordering, that I have only validated in practice that the controllers are initlialized before the traffic is allowed.

One surefire way to order this is ti inject the initializer to the V1 controllers - but let me look if there are more idiomatic options.

swaranga-netflix: Had to refactor the init service in Metacat. This was because the ApplicationRunner interface did not work as it was invoked after the Controllers and I was still getting 404s on startup. To make the order explicit, I needed to inject the initialization service to the controllers.

This also required me to split the init service to core and thrift init services to prevent a circular dependency.

ajoymajumdar: Can we use https://www.baeldung.com/spring-depends-on?

swaranga-netflix: We can. Your Spring-foo is good. Thanks. Updated.

PR #565, opened 2024-01-04
use as of system time when listing tables to improve performance

stevie9868:
**Context:**

In the context of Polaris, some database may have large number of tables, and when listing them, crdb will hold the lock and prevents other db write operations to be performed, leading to high latency.

**Blocker**
Initially, I thought we can simply include the AS OF SYSTEM in @Query for JPA, which turns out to be not the case. 
See [this commit](https://github.com/Netflix/metacat/compare/master...use_as_of_system_time_when_listing_tables) for reference. Basically it will throw `Caused by: org.postgresql.util.PSQLException: ERROR: AS OF SYSTEM TIME: only constant expressions or follower_read_timestamp are allowed`, and looks like jpa treat the passed in param as dynamic expressions.

**Alternative**
Use entityManager to execute the native query instead.

**What are some major changes**
1. Since AS OF SYSTEM only is supported in crdb, some of the unit tests, which use h2db will not work so I move those tests to functional tests which use crdb.
2. Need to create a customRepo, which hosts queries that are not feasible using JPA. In this case, the AS OF SYSTEM TIME syntax.
3. PolarisTableRepository extends the customRepo

### DIFF
--- a/metacat-connector-polaris/src/functionalTest/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableServiceFunctionalTest.java
+++ b/metacat-connector-polaris/src/functionalTest/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableServiceFunctionalTest.java
@@ -8,6 +8,7 @@ import com.netflix.metacat.common.dto.Sort;
 import com.netflix.metacat.common.dto.SortOrder;
 import com.netflix.metacat.common.server.connectors.model.TableInfo;
 import com.netflix.metacat.connector.polaris.configs.PolarisPersistenceConfig;
+import lombok.extern.slf4j.Slf4j;
 import org.junit.Assert;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
@@ -24,6 +25,7 @@ import java.util.stream.Collectors;
  * Some of the tests cannot be run in unit test as it uses h2 database, which does not support all
  * functionalities in crdb so include those tests here.
  */
+@Slf4j
 @ExtendWith(SpringExtension.class)
 @SpringBootTest(classes = {PolarisPersistenceConfig.class})
 @ActiveProfiles(profiles = {"polaris_functional_test"})
@@ -54,7 +56,7 @@ public class PolarisConnectorTableServiceFunctionalTest extends PolarisConnector
             // pause execution for 10000 milliseconds (10 seconds)
             Thread.sleep(10000);
         } catch (InterruptedException e) {
-            System.out.println("Sleep was interrupted");
+            log.debug("Sleep was interrupted");
         }
 
         List<TableInfo> tables = this.getPolarisTableService().list(
--- a/metacat-connector-polaris/src/functionalTest/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableServiceFunctionalTest.java
+++ b/metacat-connector-polaris/src/functionalTest/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableServiceFunctionalTest.java
@@ -1,0 +1,82 @@
+package com.netflix.metacat.connector.polaris;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import com.netflix.metacat.common.QualifiedName;
+import com.netflix.metacat.common.dto.Pageable;
+import com.netflix.metacat.common.dto.Sort;
+import com.netflix.metacat.common.dto.SortOrder;
+import com.netflix.metacat.common.server.connectors.model.TableInfo;
+import com.netflix.metacat.connector.polaris.configs.PolarisPersistenceConfig;
+import org.junit.Assert;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.springframework.boot.test.autoconfigure.orm.jpa.AutoConfigureDataJpa;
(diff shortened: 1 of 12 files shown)


metacat-connector-polaris/src/functionalTest/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableServiceFunctionalTest.java
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Test PolarisConnectorTableService in functional test.
+ * Some of the tests cannot be run in unit test as it uses h2 database, which does not support all
+ * functionalities in crdb so include those tests here.
+ */
+@ExtendWith(SpringExtension.class)
+@SpringBootTest(classes = {PolarisPersistenceConfig.class})
+@ActiveProfiles(profiles = {"polaris_functional_test"})
+@AutoConfigureDataJpa
+public class PolarisConnectorTableServiceFunctionalTest extends PolarisConnectorTableServiceTest {
```
rsyncu, line 31: Thanks for changing from h2db to postgresql and adding these functional tests for both the table service and store service

metacat-connector-polaris/src/functionalTest/java/com/netflix/metacat/connector/polaris/PolarisStoreConnectorFunctionalTest.java
```diff
@@ -18,4 +23,65 @@
...
+        final String tblNameA = "A_" + generateTableName();
+        final String tblNameB = "B_" + generateTableName();
+        final String tblNameC = "C_" + generateTableName();
+        createTable(dbName, tblNameA);
+        createTable(dbName, tblNameB);
+        createTable(dbName, tblNameC);
+
+        try {
+            // pause execution for 10000 milliseconds (10 seconds)
+            Thread.sleep(10000);
```
rsyncu, line 57: Is the purpose of sleep to wait for the create to complete, the create call should be synchronous and return the created object after it is done

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/store/repos/PolarisTableCustomRepositoryImpl.java
```diff
@@ -0,0 +1,78 @@
...
+        }
+        return new SliceImpl<>(resultList, page, hasNext);
+    }
+
+    @Override
+    @Transactional
+    public List<PolarisTableEntity> findAllTablesByDbNameAndTablePrefix(
+        final String dbName, final String tableNamePrefix, final int pageFetchSize) {
+        Pageable page = PageRequest.of(0, pageFetchSize, Sort.by("tbl_name").ascending());
+        entityManager.createNativeQuery("SET TRANSACTION AS OF SYSTEM TIME follower_read_timestamp()")
```
rsyncu, line 62: The `entityManager` is an instance variable in this class so if multiple threads are all running for this api at the same time, would there be situations where they overwrite the timestamp set by each other, since `follower_read_timestamp()` is a crdb function that is evaluated to a value upon execution

metacat-connector-polaris/src/functionalTest/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableServiceFunctionalTest.java
```diff
@@ -0,0 +1,82 @@
...
+        this.getPolarisTableService().create(this.getRequestContext(), tableInfo2);
+
+
+        final QualifiedName qualifiedName = QualifiedName.ofTable(CATALOG_NAME, DB_NAME, "");
+
+        try {
+            // pause execution for 10000 milliseconds (10 seconds)
+            Thread.sleep(10000);
+        } catch (InterruptedException e) {
+            System.out.println("Sleep was interrupted");
```
snehalchennuru, line 57: minor: use logger.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/store/repos/PolarisTableCustomRepositoryImpl.java
snehalchennuru: Since the entityManager is annotated with `@PersistenceContext` it's scoped within a transaction.

https://www.baeldung.com/jpa-hibernate-persistence-context

rsyncu: Thanks for the info and reference, upon closer look seems like spring will inject a special proxy for the `entityManager` in this setup so indeed should be safe
https://github.com/spring-projects/spring-framework/blob/main/spring-orm/src/main/java/org/springframework/orm/jpa/SharedEntityManagerCreator.java

metacat-connector-polaris/src/functionalTest/java/com/netflix/metacat/connector/polaris/PolarisStoreConnectorFunctionalTest.java
stevie9868: Since we are now only reading the snapshot that is follower_reader_timestamp (4.8s) ago, we need to sleep for sometime so the following list operation can read the 3 tables created.

rsyncv (dismissed):

rsyncv (approved):

PR #566, opened 2024-01-08
list table names with snapshot

stevie9868:
Also makes list table names using the AS OF SYSTEM TIME semantics.

1. Basically reuse the previous AS OF SYSTEM TIME implementation to include used case for listing table names
2. Move some unit tests to functional tests related to list table names since AS OF SYSTEM TIME is only supported in crdb

### DIFF
--- a/metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
+++ b/metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
@@ -187,7 +187,10 @@ public class PolarisConnectorTableService implements ConnectorTableService {
         try {
             final List<QualifiedName> qualifiedNames = Lists.newArrayList();
             final String tableFilter = (prefix != null && prefix.isTableDefinition()) ? prefix.getTableName() : "";
-            for (String tableName : polarisStoreService.getTables(name.getDatabaseName(), tableFilter)) {
+            for (String tableName : polarisStoreService.getTables(name.getDatabaseName(),
+                tableFilter,
+                1000)
+            ) {
                 final QualifiedName qualifiedName =
                     QualifiedName.ofTable(name.getCatalogName(), name.getDatabaseName(), tableName);
                 if (prefix != null && !qualifiedName.toString().startsWith(prefix.toString())) {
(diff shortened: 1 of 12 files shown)


metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
```diff
@@ -187,7 +187,10 @@ public List<QualifiedName> listNames(
         try {
             final List<QualifiedName> qualifiedNames = Lists.newArrayList();
             final String tableFilter = (prefix != null && prefix.isTableDefinition()) ? prefix.getTableName() : "";
-            for (String tableName : polarisStoreService.getTables(name.getDatabaseName(), tableFilter)) {
+            for (String tableName : polarisStoreService.getTables(name.getDatabaseName(),
+                tableFilter,
+                1000)
```
rsyncv, line 192: Could this constant in the code be made into a constant variable

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
```diff
@@ -388,7 +391,10 @@ public List<QualifiedName> getTableNames(
             final List<QualifiedName> result = Lists.newArrayList();
             for (int i = 0; i < databaseNames.size() && limitSize > 0; i++) {
                 final String databaseName = databaseNames.get(i);
-                final List<String> tableNames = polarisStoreService.getTables(name.getDatabaseName(), "");
+                final List<String> tableNames = polarisStoreService.getTables(
+                    name.getDatabaseName(),
+                    "",
+                    1000);
```
rsyncv, line 397: Since names are smaller than the objects themselves, it might be more efficient to make the constant larger for names, maybe we can try 5k or 10k

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/store/repos/PolarisTableCustomRepository.java
```diff
@@ -12,8 +11,9 @@ public interface PolarisTableCustomRepository {
      * @param dbName database name
      * @param tableNamePrefix table name prefix. can be empty.
      * @param pageSize target size for each page
+     * @param selectAll if true return the PolarisEntity else return name of the entity
      * @return table entities in the database.
      */
-    List<PolarisTableEntity> findAllTablesByDbNameAndTablePrefix(
-        String dbName, String tableNamePrefix, int pageSize);
+    List<?> findAllTablesByDbNameAndTablePrefix(
+        String dbName, String tableNamePrefix, int pageSize, boolean selectAll);
```
rsyncv, line 18: The `selectAll` name could be improved since the method name already has `findAll` so it could seem like it might not return the whole list, this could be renamed instead to `selectNames` or something similar

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
rsyncv: Looks like this one was missed, can you update it to read from config as well

rsyncv (approved):
Thanks for addressing the comments

PR #570, opened 2024-02-20
Include audit fields in db info

insyncoss:


stevie9868 (approved):

insyncoss:
It's actually a feature that we configure through annotations

PR #571, opened 2024-03-05
Auth Module for Connector Services

amboz:
This PR introduces the following:

1. An [authorization interface](https://github.com/Netflix/metacat/pull/571/files#diff-6e61e2ffadbf8021f520de4588b11f67523d2e4d68166fba03402b1f868db83a) for connector services to authorize requests based on `MetacatRequestContext` and return an [AuthorizationStatus](https://github.com/Netflix/metacat/pull/571/files#diff-f3657e191e22e02a134e112f448b719ddd17393d4469fb9a17201da8a280044f). [DefaultAuthorization](https://github.com/Netflix/metacat/pull/571/files#diff-4b3785d06326969f9cbb3d3562dd0d81f9c76206176b122cd65ae06a3972aee7) can be overridden as needed with custom auth logic
2. `ValidatingConnector{Catalog/Database/Partition/Table}Service` classes (e.g. [ValidatingConnectorTableService](https://github.com/Netflix/metacat/pull/571/files#diff-efa8a68ae981c19fc040f0f6c9b6c39ca5d447b47adecde0b8ac6054ef8a1ef2)) which subsume the existing throttling functionality and also include the new auth functionality
3. [ConnectorFactoryDecorator](https://github.com/Netflix/metacat/pull/571/files#diff-bfb7ff7d9fc6b444e338c9eae3c42e46eacc3135461808d406299ac3c13dc8c9) decorates connector services with validating connector services based on whether rate limiting and/or auth is enabled. Enablement of auth is based on the `connector.authorization-exempted` config ([ref](https://github.com/Netflix/metacat/pull/571/files#diff-bfb7ff7d9fc6b444e338c9eae3c42e46eacc3135461808d406299ac3c13dc8c9R159)).
4. Tests to validate: (1) [decoration and delegation for ConnectorFactoryManager](https://github.com/Netflix/metacat/pull/571/files#diff-7cd7789f181f7ac5c016b480c2ae33525f05d411ba34b31d5bf7fa5b5df03732), (2) auth and rate limiting application in validating connector services ([example](https://github.com/Netflix/metacat/pull/571/files#diff-8bea2484ddba9f755088b9dd913b23921ae3c3bc716eee744cde1ba4ccac7299)), and (3) [ConnectorManager functionality](https://github.com/Netflix/metacat/pull/571/files#diff-e0532b8ac09ff64df7bed9b80f8fc7568b68d8034781d9e2b6351a4182ace869) under all possible combinations of auth and/or rate limiting enablement.


metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/AuthEnabledConnectorTableService.java
```diff
@@ -0,0 +1,125 @@
...
+                                                          final boolean prefixSearch) {
+        return delegate.getTableNames(context, uris, prefixSearch);
+    }
+
+    @Override
+    public List<QualifiedName> getTableNames(final ConnectorRequestContext context,
+                                             final QualifiedName name,
+                                             final String filter,
+                                             final Integer limit) {
+        authorize(MetacatContextManager.getContext(), name);
```
insyncoss, line 53: Now that you're adding a new `AuthEnabledConnector` services in addition to existing `ThrottlingConnector` ones there is an opportunity for some refactor and improvement. Instead of overriding each method similar to what the throttling connector already does, there could be a parent class that does these overrides and calls an abstract `check` and both auth and throttling provide their own implementation of this.

insyncoss: That being the case, this update could also be done in a later change.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/AuthEnabledConnectorTableService.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @Override
+    public void rename(final ConnectorRequestContext context,
+                       final QualifiedName oldName,
+                       final QualifiedName newName) {
+        authorize(MetacatContextManager.getContext(), oldName);
+        delegate.rename(context, oldName, newName);
+    }
+
+    private void authorize(final MetacatRequestContext context, final QualifiedName resource) {
```
insyncoss, line 106: Adding to the previous comment, using such a framework and extending it would make it so you wouldn't need to have basically the same method and logic duplicated three times across classes which is not a great pattern for code structure or maintainability.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/AuthEnabledConnectorDatabaseService.java
```diff
@@ -0,0 +1,112 @@
...
+
+import javax.annotation.Nullable;
+import java.util.List;
+
+/**
+ * Connector that authorizes requests based on the request context.
+ */
+@Slf4j
+@RequiredArgsConstructor
+public class AuthEnabledConnectorDatabaseService implements ConnectorDatabaseService {
```
insyncoss, line 26: There is a `ConnectorBaseService` that the different types extend so having certain generalizable functionality that is needed across different types there is also an option if you see such an opportunity though up to you.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/AuthEnabledConnectorTableService.java
amboz: addressed all comments in a refactor to combine auth and throttling into one validating class 😄

metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ValidatingConnectorCatalogService.java
```diff
@@ -18,73 +22,119 @@
...
     private final RateLimiter rateLimiter;
+    private final boolean rateLimiterEnabled;
+    @NonNull
+    private final Authorization authorization;
+    private final boolean authorizationEnabled;
 
     @Override
     public void create(final ConnectorRequestContext context, final CatalogInfo resource) {
-        checkThrottling(MetacatContextManager.getContext().getRequestName(), resource.getName());
+        if (rateLimiterEnabled) {
```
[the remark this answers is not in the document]
stevie9868: For option2, maybe I miss something, but I feel like we don't need to have a config for each service.

I was thinking we still keep the two original config extempted setting (one for throttling and one for authorization), and have just 1 Validator interface and 2 Implementations (one for throttling and another for Authorization).

The interface can have one function say validate, which takes in MetacatRequestContext, and have one variable call enabled.

Then for the implementation for either throttling and Authorization, the implementation for validate should be if its enabled, then trigger the corresponding validation.

Then for each service (table/partition/catalog/database), it should have a list of Validator. I see most of the functions check both Throttling and Authorization, so we can have a generic util function that run all validators. But for some cases, where not all validators should run, we can handle them separately.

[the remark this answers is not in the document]
amboz: i see what you mean-- this is cleaner and more extensible. we will need to be careful to include the following:
1. special handling for specific methods that do not check throttling even when enabled at the connector level
2. ensure we are explicit about the order in which different validations occur. for example, always check throttling first, then auth
3. (less important) retain the ability to switch the order of each validation at the method level. this is not needed now but could be as we add more to it

will wait for others opinions before making the change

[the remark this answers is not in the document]
snehalchennuru: I prefer option 2. Not sure if Yingjian already mentioned, but we can have a Validator interface, which validates given the context. There would be 3 implementations - ThrottlingValidator, AuthorizationValidator and a ChainedValidator which takes an ordered list of Validators and executes them.  ValidatingConnectorCatalogService would take the Validator interface, but the constructor would pass in the ChainedValidator which is opinionated about the order of execution.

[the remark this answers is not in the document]
snehalchennuru: Here's a suggestion which iterates over the above suggestions:

1) As Yingjian suggested, we should have a Validator interface with two implementations—ThrottlingValidator and AuthorizationValidator—that perform validations.
2) Have a ValidationRegistry class that contains all the validators registered with it. It contains a method run that takes in a list of validations to be run. This will be used to execute the validations in the order or exact order they are to be run.
3) Implement a ValidatingConnectorBase which implements a generic ConnectorBaseService. It takes the validation registry and a delegate and implements all the methods. Each method would first perform the validation via registry.
4) Similarly, implement all the specific implementations for table, partition etc; with a similar logic.

metacat-common-server/src/test/groovy/com/netflix/metacat/common/server/connectors/ValidatingConnectorPartitionServiceSpec.groovy
```diff
@@ -0,0 +1,611 @@
...
+        try {
+            service.getPartitionNames(context, [], true)
+            throw new Success()
+        }
+        catch (Exception e) {
+            actualException = e
+        }
+
+        then:
+        if (rateLimiterEnabled || !rateLimiterEnabled) {
```
stevie9868, line 452: nit: the first if statement might be redundant

metacat-common-server/src/test/groovy/com/netflix/metacat/common/server/connectors/ValidatingConnectorTableServiceSpec.groovy
```diff
@@ -0,0 +1,514 @@
...
+        try {
+            service.getTableNames(context, [], true)
+            throw new Success()
+        }
+        catch (Exception e) {
+            actualException = e
+        }
+
+        then:
+        if (rateLimiterEnabled || !rateLimiterEnabled) {
```
stevie9868, line 447: nit: this first if statement might be redundant

stevie9868:
Looks like previously, there are two places where we don't checkThrottling 

ThrottlingConnectorPartitionService getPartitionNames
and
ThrottlingConnectorTableService.java getTableNames

Do we know why these two calls can be exempted from throttling check?

amboz:
> Looks like previously, there are two places where we don't checkThrottling
> 
> ThrottlingConnectorPartitionService getPartitionNames and ThrottlingConnectorTableService.java getTableNames
> 
> Do we know why these two calls can be exempted from throttling check?

@stevie9868 my undersanding these methods dont have a qualified resource associated with them, just prefix searches, and rate limiting is currently only on a per-resource basis

metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorFactoryDecorator.java
```diff
@@ -32,24 +36,33 @@ public ConnectorFactoryDecorator(@NonNull final ConnectorPlugin connectorPlugin,
...
+                    + "plugin-type: {}, catalog: {}, shard: {}",
                 connectorContext.getConnectorType(), connectorPlugin.getType(),
                 connectorContext.getCatalogName(), connectorContext.getCatalogShardName());
-            service = new ThrottlingConnectorCatalogService(service, rateLimiter);
+            service = new ValidatingConnectorCatalogService(
+                service,
+                rateLimiter,
+                rateLimiterEnabled,
+                authorization,
+                authorizationEnabled
```
akashdw, line 64: The authorization object should have the capability to determine its own enabled status, rather than having it flagged externally.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorFactoryDecorator.java
```diff
@@ -74,12 +93,18 @@ public ConnectorDatabaseService getDatabaseService() {
...
+                    + "plugin-type: {}, catalog: {}, shard: {}",
                 connectorContext.getConnectorType(), connectorPlugin.getType(),
                 connectorContext.getCatalogName(), connectorContext.getCatalogShardName());
-            service = new ThrottlingConnectorTableService(service, rateLimiter);
+            service = new ValidatingConnectorTableService(
+                service,
+                rateLimiter,
+                rateLimiterEnabled,
+                authorization,
+                authorizationEnabled
```
akashdw, line 106: The authorization object should have the capability to determine its own enabled status, rather than having it flagged externally.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorFactoryDecorator.java
```diff
@@ -89,12 +114,18 @@ public ConnectorTableService getTableService() {
...
+                    + "plugin-type: {}, catalog: {}, shard: {}",
                 connectorContext.getConnectorType(), connectorPlugin.getType(),
                 connectorContext.getCatalogName(), connectorContext.getCatalogShardName());
-            service = new ThrottlingConnectorPartitionService(service, rateLimiter);
+            service = new ValidatingConnectorPartitionService(
+                service,
+                rateLimiter,
+                rateLimiterEnabled,
+                authorization,
+                authorizationEnabled
```
akashdw, line 127: The authorization object should have the capability to determine its own enabled status, rather than having it flagged externally.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ValidatingConnectorCatalogService.java
```diff
@@ -18,73 +22,119 @@
...
         delegate.update(context, resource);
     }
 
     @Override
     public void delete(final ConnectorRequestContext context, final QualifiedName name) {
-        checkThrottling(MetacatContextManager.getContext().getRequestName(), name);
+        if (rateLimiterEnabled) {
+            checkThrottling(MetacatContextManager.getContext().getRequestName(), name);
+        }
+        if (authorizationEnabled) {
```
akashdw, line 67: The responsibility for this and similar `if` blocks should lie with the Authorization implementation, rather than the caller performing the checks.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/api/authorization/Authorization.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+public interface Authorization {
+
+    /**
+     * Authorize a given request based on the request context.
+     *
+     * @param context The request context.
+     * @return True if the request is authorized and False if unauthorized.
+     */
+    default AuthorizationStatus isAuthorized(@NonNull MetacatRequestContext context) {
```
akashdw, line 36: The interface could be more explicit about its expectations from the MetacatRequestContext, instead of passing down the entire context.

akashdw (changes requested):

PR #572, opened 2024-03-06
Fix setTag logic to prevent high cpu load on rds

stevie9868:
For now, whenever we want to set a tag on a resource, it will always fetch all tags from the dbs into memory, and then do a diff to figure out what tags never exist in the db. Loading all tags at once creates a lot of load on the rds if the number of tags is huge when number of setTag request is high. 

For this PR, I also remove some unused code.

Instead of fetching all tags into memory, we will instead do 
```
"INSERT INTO lookup_values(lookup_id, values_string) VALUES (?,?) "
            + "ON DUPLICATE KEY UPDATE lookup_id=lookup_id, values_string=values_string"; 
```

Even though, there is no unit test for tag component, there are lots of existing tests in the functional test, and I also add one more functional test to cover more scenarios.
Since the logic I change should have no impact on the core logic, we should expect the functionalTest to have the same result [before](https://github.com/Netflix/metacat/pull/573) and after this change.

Finally, in order to make this insert query fast, we will need to build a composite index on (lookup_id, values_string)

### DIFF
--- a/metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlLookupService.java
+++ b/metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlLookupService.java
@@ -50,8 +50,7 @@ public class MySqlLookupService implements LookupService {
     private static final String SQL_INSERT_LOOKUP_VALUES =
         "insert into lookup_values( lookup_id, values_string) values (?,?)";
     private static final String SQL_INSERT_LOOKUP_VALUE_IF_NOT_EXIST =
-        "INSERT INTO lookup_values(lookup_id, values_string) VALUES (?,?) "
-            + "ON DUPLICATE KEY UPDATE lookup_id=lookup_id, values_string=values_string";
+        "INSERT IGNORE INTO lookup_values (lookup_id, values_string) VALUES (?, ?)";
 
     private static final String SQL_GET_LOOKUP_VALUES =
         "select values_string value from lookup_values where lookup_id=?";
--- a/metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlLookupService.java
+++ b/metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlLookupService.java
@@ -13,7 +13,6 @@
 
 package com.netflix.metacat.metadata.mysql;
 
-import com.google.common.base.Joiner;
 import com.google.common.collect.Sets;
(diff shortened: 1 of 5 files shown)


metacat-metadata-mysql/src/test/groovy/com/netflix/metacat/metadata/mysql/MySqlLookupServiceSpec.groovy
```diff
@@ -0,0 +1,79 @@
...
+
+    def setup() {
+        config = Mock(Config)
+        jdbcTemplate = Mock(JdbcTemplate)
+
+        mySqlLookupService = new MySqlLookupService(config, jdbcTemplate)
+    }
+
+    @Unroll
+    def "test includeValues"() {
```
insyncoss, line 48: Can you add some additional tests featuring duplicates and multiple call iterations

stevie9868: Sorry, could you elaborate more?

I think I have covered them also in the functional test.

insyncoss: ok looks like the functional test can cover the iterations part, could you add another unit test featuring if a caller tries to add duplicate tag in one call like pass in the same tag multiple times in the input list

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlLookupService.java
```diff
@@ -244,20 +206,14 @@ private Lookup findOrCreateLookupByName(final String name) throws SQLException {
...
-                lookup.setValues(values);
-            } else {
-                inserts = Sets.difference(values, lookupValues);
+            final Lookup lookup = findOrCreateLookupByName(name, includeValues);
+            if (!values.isEmpty()) {
+                insertLookupValuesIfNotExist(lookup.getId(), values);
             }
-            if (!inserts.isEmpty()) {
-                insertLookupValues(lookup.getId(), inserts);
+            if (includeValues) {
```
insyncoss, line 215: the previous code didn't even add back the new tags to the return result, but based on usage it does not make a difference either way

stevie9868: yep, you are right but I want to make it consistent.

insyncoss (approved):

PR #589, opened 2024-04-26
Yingjianw/list db with follower reader timestamp

stevie9868:
As we see more dbs added to our metadata service, the load on crdb increases when there are many listDB calls, and we see only subset of certain nodes in our cluster experiencing high cpu usage.

Thus, in order to better spread the traffic, for list dbs operation, we are going to use the read_follower_timestamp.

Unfortunately, we need to rollback previous 2 commits as we don't want them to be included into the same rollout

### DIFF
--- a/metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorDatabaseService.java
+++ b/metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorDatabaseService.java
@@ -18,7 +18,6 @@ import com.netflix.metacat.connector.polaris.store.PolarisStoreService;
 import com.netflix.metacat.connector.polaris.store.entities.PolarisDatabaseEntity;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.dao.DataIntegrityViolationException;
-import org.springframework.transaction.annotation.Transactional;
 
 import javax.annotation.Nullable;
 import java.util.ArrayList;
@@ -35,6 +34,8 @@ public class PolarisConnectorDatabaseService implements ConnectorDatabaseService
     private final String defaultLocationPrefix;
     private final PolarisStoreService polarisStoreService;
 
+    private final ConnectorContext connectorContext;
+
     /**
      * Constructor.
      *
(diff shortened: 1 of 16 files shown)


insyncoss:
To make things cleaner, could you first separate out the revert changes from the core logic changes?

metacat-functional-tests/src/functionalTest/groovy/com/netflix/metacat/MetacatFunctionalSpec.groovy
```diff
@@ -226,6 +226,7 @@ class MetacatFunctionalSpec extends Specification {
 
         when:
         api.createDatabase(catalog.name, databaseName, dto)
+        Thread.sleep(5000)
```
insyncoss, line 229: In the other tests the wait time is 10 seconds delay is this delay enough for the follower read?

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorDatabaseService.java
```diff
@@ -159,17 +160,12 @@ public List<QualifiedName> listNames(
         @Nullable final Pageable pageable
     ) {
         try {
-            List<QualifiedName> qualifiedNames = polarisStoreService.getAllDatabases().stream()
-                .map(d -> QualifiedName.ofDatabase(name.getCatalogName(), d.getDbName()))
+            final String dbPrefix = prefix == null ? "" : prefix.getDatabaseName();
+            final List<QualifiedName> qualifiedNames = polarisStoreService.getDatabaseNames(
+                dbPrefix, sort, 1000)
```
insyncoss, line 165: Should we try a larger page size like 5000 or 10000? That's what we bumped it to for table names and it seems it could be more efficient.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorDatabaseService.java
```diff
@@ -77,6 +77,7 @@ public void create(final ConnectorRequestContext context, final DatabaseInfo dat
      * {@inheritDoc}.
      */
     @Override
+    @Transactional
```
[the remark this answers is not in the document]
stevie9868: Good find. Not sure why I added this, might be something I accidentally included during my rebase...

metacat-functional-tests/src/functionalTest/groovy/com/netflix/metacat/MetacatFunctionalSpec.groovy
stevie9868: yeah, it turns out 5 seconds is enough. I will adjust other test to use 5 seconds as well.

insyncoss (approved):

PR #593, opened 2024-05-07
Publish new name on table rename event

insyncoss:
Publish new name on table rename event, and add publish of new name in case of SNS message send failure as well. This allows clients to properly be notified of new name after rename event even in case of initial send failure.

### DIFF
--- a/metacat-common/src/main/java/com/netflix/metacat/common/dto/notifications/sns/messages/RenameTableMessage.java
+++ b/metacat-common/src/main/java/com/netflix/metacat/common/dto/notifications/sns/messages/RenameTableMessage.java
@@ -26,6 +26,8 @@ import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.ToString;
 
+import javax.annotation.Nullable;
+
 /**
  * A message sent when a table is renamed.
  *
@@ -35,6 +37,7 @@ import lombok.ToString;
 @ToString(callSuper = true)
 @EqualsAndHashCode(callSuper = true)
 public class RenameTableMessage extends UpdateOrRenameTableMessageBase {
+    private final String newName;
 
     /**
      * Create a new RenameTableMessage.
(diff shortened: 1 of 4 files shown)


snehalchennuru (approved):
lgtm

metacat-common/src/main/java/com/netflix/metacat/common/dto/notifications/sns/messages/RenameTableMessage.java
```diff
@@ -51,8 +55,10 @@ public RenameTableMessage(
         @JsonProperty("timestamp") final long timestamp,
         @JsonProperty("requestId") final String requestId,
         @JsonProperty("name") final String name,
-        @JsonProperty("payload") final UpdatePayload<TableDto> payload
+        @JsonProperty("newName") final String newName,
+        @Nullable @JsonProperty("payload") final UpdatePayload<TableDto> payload
```
stevie9868, line 59: Wondering why do we want to have the payload be Nullable?

stevie9868 (approved):

amboz (approved):

metacat-common/src/main/java/com/netflix/metacat/common/dto/notifications/sns/messages/RenameTableMessage.java
insyncoss: In case the initial send fails, for example if full payload causes message to be too long, then the fallback is send the message again except null payload.

metacat-main/src/main/java/com/netflix/metacat/main/services/notifications/sns/SNSNotificationServiceImpl.java
```diff
@@ -349,11 +349,13 @@ private UpdateOrRenameTableMessageBase createUpdateorRenameTableMessage(
                     new UpdatePayload<>(oldTable, patch)
                 );
             } else {
+                final QualifiedName newName = currentTable == null ? null : currentTable.getName();
                 return new RenameTableMessage(
                     id,
                     timestamp,
                     requestId,
                     name.toString(),
+                    newName == null ? "" : newName.toString(),
```
ajoymajumdar, line 358: Why not set it to null if null?

insyncoss: Setting to null instead of empty string if null would be fine as well. If defaulting to null instead of empty string here helps on the client side then we can make that change.

PR #595, opened 2024-06-02
add ParentChildRelationSqlService

stevie9868:
### PR Summary: Introduction of Parent-Child Relationship Metadata in Metacat

This pull request introduces metadata to manage parent-child relationships within Metacat. The key invariants for these relationships are:

- **Child Table**: Once created, a child table’s parent cannot be altered unless the parent itself is renamed.
- **Parent Table**: Modifications to a parent table’s list of children are restricted to:
  - Adding a new clone table.
  - Renaming or dropping an existing child clone table.

The implementation is integrated into the open-source Metacat project, as parent-child relationships will be a fundamental concept, especially for tables like Iceberg.

#### Important Classes:
1. **MySQLParentChildRelMetadataService.java**: 
   - Provides CRUD services to manipulate the parent-child relationship metadata.
   
2. **TableServiceImpl**: 
   - Integrates `MySQLParentChildRelMetadataService` into the table management workflow.

   **Operations**:

   - **Create**:
     - The parent-child relationship is always established before the table is created. If table creation fails, the parent-child relationship is then deleted.
     - This order has two advantages:
       1. Reverting changes to the parent-child relationship is less destructive.
       2. Since both the table name and UUID are stored, it is safe to delete the record if necessary.

   - **Rename**:
     - Similar to the create operation, the parent-child relationship is renamed before the table is renamed. And if table renamed fail, we will try to rename the parent-child relation back.

   - **Drop**:
     - For the drop operation, the table is always dropped first, followed by the parent-child relationship if applicable.
     - If the parent-child relationship cannot be dropped, it will act as a dangling reference.
   
   - **Get**:
     - We will always fetch the parent child relation from `MySQLParentChildRelMetadataService` and then insert it into the definitionMetadata. For now we will only populate the field for child table. For parent to get the list of children, we expose a search endpoint.
   
3. **metacat.sql**:
   - Define the schema of the parent child relation schema

### DIFF
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/model/BaseRelEntityInfo.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/model/BaseRelEntityInfo.java
@@ -1,6 +1,8 @@
 package com.netflix.metacat.common.server.model;
 
+import lombok.AllArgsConstructor;
 import lombok.Data;
+import lombok.NoArgsConstructor;
 
 import java.io.Serializable;
 
@@ -8,28 +10,12 @@ import java.io.Serializable;
  * Base class to represent relation entity.
  */
 @Data
+@AllArgsConstructor
+@NoArgsConstructor
 public abstract class BaseRelEntityInfo implements Serializable {
     private static final long serialVersionUID = 9121109874202888889L;
     private String name;
     private String relationType;
     private String uuid;
 
-    /**
-     Empty Constructor.
-     */
-    public BaseRelEntityInfo() {
-
-    }
-
-    /**
-     Constructor with all params.
-     @param name name of the entity
-     @param relationType type of the relation
-     @param uuid uuid of the entity
-     */
-    public BaseRelEntityInfo(final String name, final String relationType, final String uuid) {
-        this.name = name;
-        this.relationType = relationType;
-        this.uuid = uuid;
-    }
 }
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/model/BaseRelEntityInfo.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/model/BaseRelEntityInfo.java
@@ -1,0 +1,32 @@
+package com.netflix.metacat.common.server.model;
+
+import lombok.Data;
+
+/**
+ * Base class to represent relation entity.
+ */
+@Data
+public abstract class BaseRelEntityInfo {
+    private String name;
+    private String relationType;
+    private String uuid;
+
+    /**
+     Empty Constructor.
+     */
+    public BaseRelEntityInfo() {
+
+    }
+
+    /**
+     Constructor with all params.
+     @param name name of the entity
+     @param relationType type of the relation
+     @param uuid uuid of the entity
+     */
+    public BaseRelEntityInfo(final String name, final String relationType, final String uuid) {
+        this.name = name;
+        this.relationType = relationType;
+        this.uuid = uuid;
+    }
+}
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/model/BaseRelEntityInfo.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/model/BaseRelEntityInfo.java
@@ -2,11 +2,14 @@ package com.netflix.metacat.common.server.model;
 
 import lombok.Data;
 
+import java.io.Serializable;
+
 /**
  * Base class to represent relation entity.
  */
 @Data
-public abstract class BaseRelEntityInfo {
+public abstract class BaseRelEntityInfo implements Serializable {
+    private static final long serialVersionUID = 9121109874202888889L;
     private String name;
     private String relationType;
     private String uuid;
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/model/ChildInfo.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/model/ChildInfo.java
@@ -1,0 +1,23 @@
+package com.netflix.metacat.common.server.model;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+/**
+ * ChildInfo.
+ */
+@EqualsAndHashCode(callSuper = true)
+@AllArgsConstructor
+@Data
+public class ChildInfo extends BaseRelEntityInfo {
+    /**
+     Constructor with all params.
+     @param name name of the entity
+     @param relationType type of the relation
+     @param uuid uuid of the entity
+     */
+    public ChildInfo(final String name, final String relationType, final String uuid) {
+        super(name, relationType, uuid);
+    }
+}
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/usermetadata/ParentChildRelMetadataConstants.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/usermetadata/ParentChildRelMetadataConstants.java
@@ -7,37 +7,46 @@ package com.netflix.metacat.common.server.usermetadata;
  */
 public final class ParentChildRelMetadataConstants {
     /**
-     * During create, top level key specified in DefinitionMetadata that indicate the parent table name.
+     * During get and create, top level key specified in DefinitionMetadata that indicates the parent child infos.
      */
-    public static final String PARENTNAME = "root_table_name";
+    public static final String PARENT_CHILD_RELINFO = "parentChildRelationInfo";
     /**
-     * During create, top level key specified in DefinitionMetadata that indicates the parent table uuid.
+     * During create, nested level key specified in DefinitionMetadata['parentChildRelationInfo']
+     * that indicate the parent table name.
      */
-    public static final String PARENTUUID = "root_table_uuid";
+    public static final String PARENT_NAME = "root_table_name";
     /**
-     * During create, top level key specified in DefinitionMetadata that indicates the child table uuid.
+     * During create, nested level key specified in DefinitionMetadata['parentChildRelationInfo']
+     * that indicates the parent table uuid.
      */
-    public static final String CHILDUUID = "child_table_uuid";
+    public static final String PARENT_UUID = "root_table_uuid";
+    /**
+     * During create, nested level key specified in DefinitionMetadata['parentChildRelationInfo']
+     * that indicates the child table uuid.
+     */
+    public static final String CHILD_UUID = "child_table_uuid";
 
     /**
-     * During create, top level key specified in DefinitionMetadata that indicates relationType.
+     * During create, nested level key specified in DefinitionMetadata['parentChildRelationInfo']
+     * that indicates relationType.
      */
-    public static final String RELATIONTYPE = "relationType";
-
-    /**
-     * During get, top level key specified in DefinitionMetadata that indicates the parent child infos.
-     */
-    public static final String PARENTCHILDRELINFO = "parentChildRelationInfo";
+    public static final String RELATION_TYPE = "relationType";
 
     /**
      * During get, the nested key specified in DefinitionMetadata[PARENTCHILDRELINFO] that indicates parent infos.
      */
-    public static final String PARENTINFOS = "parentInfos";
+    public static final String PARENT_INFOS = "parentInfos";
 
     /**
      * During get, the nested key specified in DefinitionMetadata[PARENTCHILDRELINFO] that indicates child infos.
      */
-    public static final String CHILDINFOS = "childInfos";
+    public static final String CHILD_INFOS = "childInfos";
+
+    /**
+     * During get, the nested key specified in DefinitionMetadata[PARENTCHILDRELINFO]
+     * that indicates if a table is parent.
+     */
+    public static final String IS_PARENT = "isParent";
 
     private ParentChildRelMetadataConstants() {
 
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/usermetadata/ParentChildRelMetadataConstants.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/usermetadata/ParentChildRelMetadataConstants.java
@@ -1,0 +1,46 @@
+package com.netflix.metacat.common.server.usermetadata;
+
+/**
+ * ParentChildRelMetadataConstants.
+ *
+ * @author yingjianw
+ */
+public final class ParentChildRelMetadataConstants {
+    /**
+     * During create, top level key specified in DefinitionMetadata that indicate the parent table name.
+     */
+    public static final String PARENTNAME = "root_table_name";
+    /**
+     * During create, top level key specified in DefinitionMetadata that indicates the parent table uuid.
+     */
+    public static final String PARENTUUID = "root_table_uuid";
+    /**
+     * During create, top level key specified in DefinitionMetadata that indicates the child table uuid.
+     */
+    public static final String CHILDUUID = "child_table_uuid";
+
+    /**
+     * During create, top level key specified in DefinitionMetadata that indicates relationType.
+     */
+    public static final String RELATIONTYPE = "relationType";
+
+    /**
+     * During get, top level key specified in DefinitionMetadata that indicates the parent child infos.
+     */
+    public static final String PARENTCHILDRELINFO = "parentChildRelationInfo";
+
+    /**
+     * During get, the nested key specified in DefinitionMetadata[PARENTCHILDRELINFO] that indicates parent infos.
+     */
+    public static final String PARENTINFOS = "parentInfos";
+
+    /**
+     * During get, the nested key specified in DefinitionMetadata[PARENTCHILDRELINFO] that indicates child infos.
+     */
+    public static final String CHILDINFOS = "childInfos";
+
+    private ParentChildRelMetadataConstants() {
+
+    }
+
+}
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/usermetadata/ParentChildRelMetadataService.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/usermetadata/ParentChildRelMetadataService.java
@@ -1,5 +1,6 @@
 package com.netflix.metacat.common.server.usermetadata;
 import com.netflix.metacat.common.QualifiedName;
+import com.netflix.metacat.common.dto.notifications.ChildInfoDto;
 import com.netflix.metacat.common.server.model.ChildInfo;
 import com.netflix.metacat.common.server.model.ParentInfo;
 
@@ -91,6 +92,15 @@ public interface ParentChildRelMetadataService {
      * @return a set of ChildInfo
      */
     Set<ChildInfo> getChildren(
+        QualifiedName name
+    );
+
+    /**
+     * get the set of children dto for the input name.
+     * @param name name
+     * @return a set of ChildInfo
+     */
+    Set<ChildInfoDto> getChildrenDto(
         QualifiedName name
     );
 }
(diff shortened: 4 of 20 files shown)


metacat-common-server/src/main/java/com/netflix/metacat/common/server/model/ChildInfo.java
```diff
@@ -0,0 +1,16 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+/**
+ * ChildInfo.
+ */
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class ChildInfo {
```
stevie9868, line 13: Probably a cleaner interface to have a base RelationInfo Class with name and relationType as members and have ChildInfo and parentInfo extends it

metacat-common-server/src/main/java/com/netflix/metacat/common/server/usermetadata/ParentChildRelMetadataService.java
```diff
@@ -0,0 +1,78 @@
...
+        QualifiedName name
+    );
+
+    /**
+     * get the parent for the input name.
+     *
+     * @param name  name
+     * @return parentInfo
+     */
+    ParentInfo getParent(
```
stevie9868, line 65: Should we return a Set of ParentInfo even though we only have one parent right now?

Downstream services that depends on this field can implement with the assumption that the parent filed will always be a set and they don't need to worry about changing their implementations once we support multi-parents

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -270,7 +353,19 @@ public TableDto deleteAndReturn(final QualifiedName name, final boolean isMView)
         // Try to delete the table even if get above fails
         try {
             connectorTableServiceProxy.delete(name);
```
stevie9868, line 355: For drop, do we want to have `parentChildRelMetadataService.drop` run before `connectorTableServiceProxy.delete(name)`?

There are two pros for running `parentChildRelMetadataService.drop` first:
1. Validation that prevents dropping the children is actually performed in `parentChildRelMetadataService.drop`, and normally it makes sense to perform validation before taking certain action.
2. Dropping and then recover a table if there is any failures seems more destructive than dropping the parent child relation and then recovering the parent child relation.

Cons:
1. Race condition, between `parentChildRelMetadataService.drop` and `connectorTableServiceProxy.delete(name)`, a new clone table is created

Or we can always have a validation (read the children list and make sure there are no children) before `parentChildRelMetadataService.drop`, and then still keep  validation in `parentChildRelMetadataService.drop` as a safety guard.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -109,6 +115,7 @@ public TableDto create(final QualifiedName name, final TableDto tableDto) {
         connectorTableServiceProxy.create(name, converterUtil.fromTableDto(tableDto));
```
stevie9868, line 115: Should we run `saveParentChildRelationship(name, tableDto)` before `connectorTableServiceProxy.create`?

Pros:
1. To prevent read after write inconsistency, all the validations are done in mysql level, which are in `saveParentChildRelationship(name, tableDto)`. For now if the validation failed, we will need to drop the table as the table is already created. On the other hand, if we switch the order, if validation failed, we don't need to delete the table as it is not created.
2. For now, if there is any failure when saving parent child relation, we need to delete the created table. If we switch the order, if we fail to create the table, we just need to drop the parent child relation.

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlParentChildRelMetaDataService.java
```diff
@@ -141,8 +141,8 @@ public void drop(final QualifiedName name) {
                     return ps;
                 });
 
-                if (count == 0) {
-                    throw new RuntimeException("Fail to drop " + name.toString() + " because it still has child table");
+                if (count == 0 && getChildren(name).size() != 0) {
```
stevie9868, line 144: Need to have `getChildren(name).size() != 0` because majority of the tables don ot have child table and when we try to execute the drop query, the count will always be zero. And the validation should only fail those that still have parents

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlParentChildRelMetaDataService.java
```diff
@@ -0,0 +1,182 @@
...
+ * This stores the parent child relationship of two entities as first class citizen in Metacat.
+ */
+@Slf4j
+@SuppressFBWarnings
+@Transactional("metadataTxManager")
+public class MySqlParentChildRelMetaDataService implements ParentChildRelMetadataService {
+    // For now,
+    // First where clause ? = parent : Prevent the table that is a child for another to be created as a parent
+    // Second where clause ? = child : Prevent the child having more than one parent
+    static final String SQL_CREATE_PARENT_CHILD_RELATIONS =
```
stevie9868, line 31: For read committed level, which can still have phatom read, we can still have two concurrent transaction pass the exist call and end up with invalid result

Given the limitation, probably we can separate the validation out of this SQL?

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlParentChildRelMetaDataService.java
```diff
@@ -0,0 +1,182 @@
...
+            + "WHERE NOT EXISTS (SELECT 1 FROM parent_child_relation WHERE child = ?) "
+            + "AND NOT EXISTS (SELECT 1 FROM parent_child_relation WHERE child = ?)";
+
+    static final String SQL_RENAME_PARENT_ENTITY = "UPDATE parent_child_relation "
+        + "SET parent = ? WHERE parent = ?";
+    static final String SQL_RENAME_CHILD_ENTITY = "UPDATE parent_child_relation "
+        + "SET child = ? WHERE child = ?";
+
+    static final String SQL_DROP_ENTITY = "DELETE FROM parent_child_relation "
+            + "WHERE child = ? "
```
stevie9868, line 43: Same as create, it will also have similar issue.

stevie9868:
Another thing to discuss is should we put validation along with with mysql validation. 

Since we are using read_committed, we may still face phantom read problem so having the validation inside the mysql query also does not help.

metacat-functional-tests/metacat-test-cluster/datastores/mysql/docker-entrypoint-initdb.d/metacat.sql
```diff
@@ -44,6 +44,38 @@ CREATE TABLE data_metadata_delete (
...
+                                       `parent_uuid` varchar(255) NOT NULL,
+                                       `child` varchar(255) NOT NULL,
+                                       `child_uuid` varchar(255) NOT NULL,
+                                       `relation_type` varchar(255) NOT NULL,
+                                       PRIMARY KEY (`parent`, `child`, `parent_uuid`, `child_uuid`, `relation_type`),
+                                       INDEX `idx_child` (`child`)
+) ENGINE=InnoDB AUTO_INCREMENT=10078235 DEFAULT CHARSET=latin1;
+/*!40101 SET character_set_client = @saved_cs_client */;
+
+DROP TABLE IF EXISTS `definition_metadata`;
```
stevie9868, line 62: accidentally add this for definition_metadata , will remove this in the later PR.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/model/BaseRelEntityInfo.java
```diff
@@ -0,0 +1,32 @@
+package com.netflix.metacat.common.server.model;
+
+import lombok.Data;
+
+/**
+ * ChildInfo.
```
jfz, line 6: typo

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -139,6 +153,86 @@ public TableDto create(final QualifiedName name, final TableDto tableDto) {
...
+        if (tableDto.getDefinitionMetadata() != null) {
+            final ObjectNode definitionMetadata = tableDto.getDefinitionMetadata();
+            if (definitionMetadata.has(ParentChildRelMetadataConstants.PARENTNAME)) {
+                // fetch parent name
+                final String parentFullName = definitionMetadata.path(ParentChildRelMetadataConstants.PARENTNAME)
+                    .asText();
+                String parentUUID;
+                String childUUID;
+
+                if (!definitionMetadata.has(ParentChildRelMetadataConstants.PARENTUUID)) {
```
stevie9868, line 186: I plan to add this validation internally

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -373,11 +474,24 @@ public Optional<TableDto> get(final QualifiedName name, final GetTableServicePar
...
 
         if (getTableServiceParameters.isIncludeDefinitionMetadata()) {
-            final Optional<ObjectNode> definitionMetadata =
+            Optional<ObjectNode> definitionMetadata =
                 (getTableServiceParameters.isDisableOnReadMetadataIntercetor())
                     ? userMetadataService.getDefinitionMetadata(name)
                     : userMetadataService.getDefinitionMetadataWithInterceptor(name,
                     GetMetadataInterceptorParameters.builder().hasMetadata(tableInternal).build());
+            // Always get the source of truth for parent child relation from the parentChildRelMetadataService
+            final Set<ParentInfo> parentInfo = parentChildRelMetadataService.getParents(name);
```
amboz, line 483: if getParents() throws an exception, are we comfortable failing the whole getTable call? i believe that is the correct thing to do, but want to call out that this has effects on the majority of tables despite them having no parent-child relationship

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -437,7 +551,24 @@ public void rename(
         if (oldTable != null) {
             //Ignore if the operation is not supported, so that we can at least go ahead and save the user metadata
             eventBus.post(new MetacatRenameTablePreEvent(oldName, metacatRequestContext, this, newName));
-            connectorTableServiceProxy.rename(oldName, newName, isMView);
+
+            // Before rename, first rename its parent child relation
```
amboz, line 555: making a note this will change if we make that adjustment to duplicate at first step of rename as discussed offline

amboz: we can also utilize the callback structure you have above for rolling back the create

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlParentChildRelMetaDataService.java
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    @Override
+    public void createParentChildRelation(final QualifiedName parentName,
+                                          final String parentUUID,
+                                          final QualifiedName childName,
+                                          final String childUUID,
+                                          final String type) {
+        // Validation to prevent having a child have two parents
+        final Set<ParentInfo> childParents = getParents(childName);
```
amboz, line 76: we can move this enforcement to the db with:
```
CREATE TABLE `parent_child_relation` (
    `parent` varchar(255) NOT NULL,
    `parent_uuid` varchar(255) NOT NULL,
    `child` varchar(255) NOT NULL,
    `child_uuid` varchar(255) NOT NULL,
    `relation_type` varchar(255) NOT NULL,
    PRIMARY KEY (`parent`, `child`, `parent_uuid`, `child_uuid`, `relation_type`),
    UNIQUE KEY `unique_child` (`child`, `child_uuid`),
    INDEX `idx_child` (`child`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1;
```

but this would break our new rename logic where we have that temporary stage with both old and new relationship. we need to handle that either way in this method, so perhaps we dont enforce at the db and we pass down a `isRename` arg to this method

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlParentChildRelMetaDataService.java
```diff
@@ -0,0 +1,229 @@
...
+        jdbcTemplate.update(connection -> {
+            final PreparedStatement ps = connection.prepareStatement(SQL_RENAME_CHILD_ENTITY);
+            ps.setString(1, newName.toString());
+            ps.setString(2, oldName.toString());
+            return ps;
+        });
+    }
+
+    @Override
+    public void drop(final QualifiedName name) {
```
amboz, line 172: nit: we could combine into one sql (delete from ___ where parent = ? or child = ?), but its not necessarily a safer option since this class has a txn manager on it

amboz (dismissed):
+1 with some minor comments. however, ping me again if we incorporate more changes per offline convos :]

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -437,7 +551,24 @@ public void rename(
...
+            parentChildRelMetadataService.rename(oldName, newName);
+
+            try {
+                connectorTableServiceProxy.rename(oldName, newName, isMView);
+            } catch (Exception e) {
+                try {
+                    // if rename operation fail, rename back the parent child relation
+                    parentChildRelMetadataService.rename(newName, oldName);
```
insyncoss, line 563: If making the two stage commit changes, in first step insert a relation with the new name, then in this step if exception thrown delete the relation with the new name.

If no exception then delete the relation with the old name.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -139,6 +153,86 @@ public TableDto create(final QualifiedName name, final TableDto tableDto) {
...
+                    throw new RuntimeException("root_table_uuid is not specified for parent table=" + parentFullName);
+                }
+
+                if (!definitionMetadata.has(ParentChildRelMetadataConstants.CHILDUUID)) {
+                    throw new RuntimeException("child_table_uuid is not specified for child table=" + child);
+                }
+                parentUUID = definitionMetadata.path(ParentChildRelMetadataConstants.PARENTUUID).asText();
+                childUUID = definitionMetadata.path(ParentChildRelMetadataConstants.CHILDUUID).asText();
+
+                final String[] splits = parentFullName.split("\\.");
```
insyncoss, line 196: What's the reason to switch between the qualified name string using '/' as delimiter vs this way? Unifying on a single pattern like using '/' would be more uniform and not require conversions.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -267,10 +361,12 @@ public TableDto deleteAndReturn(final QualifiedName name, final boolean isMView)
             }
         }
 
-        // Try to delete the table even if get above fails
+        final Set<ChildInfo> childInfos = parentChildRelMetadataService.getChildren(name);
+        if (childInfos != null && !childInfos.isEmpty()) {
+            throw new RuntimeException("Fail to drop " + name + " because it still has child table");
```
insyncoss, line 366: It might help to print at least one of the child tables in the exception message.

Since the child tables are not included in get table response, and most users will not know how to use the special get child endpoint. So if they try to drop a table and it fails then the exception message can at least tell them which of the child tables exist that is blocking the drop.

insyncoss (approved):

metacat-functional-tests/src/functionalTest/groovy/com/netflix/metacat/MetacatSmokeSpec.groovy
```diff
@@ -1962,4 +1966,304 @@ class MetacatSmokeSpec extends Specification {
...
+         */
+        when:
+        api.renameTable(catalogName, databaseName, parent1, renameParent1)
+        parent1Table = api.getTable(catalogName, databaseName, renameParent1, true, true, false)
+        child11Table = api.getTable(catalogName, databaseName, child11, true, true, false)
+        child11ParentChildRelationInfo = child11Table.definitionMetadata.get("parentChildRelationInfo")
+        child12Table = api.getTable(catalogName, databaseName, child12, true, true, false)
+        child12ParentChildRelationInfo = child12Table.definitionMetadata.get("parentChildRelationInfo")
+
+        then:
```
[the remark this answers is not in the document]
stevie9868: It's not easy to simulate such scenario in this e2e smoke test. I plan to have a equivalent one on parentChildService e2e test

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
stevie9868: For now, when getDefinitionMetadata failed, we will also throw an exception so I think it's ok.

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlParentChildRelMetaDataService.java
stevie9868: I agree but I decide to separate them out to anticipate in the future where we need to only drop one of them

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlParentChildRelMetaDataService.java
stevie9868: I agree but then the error message we populate might not be clear because it might break the primary key unique or the unique key and if we define them in db level, it's hard to tell which key was violated.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
stevie9868: good point, our internal client actually will pass a . separated. Probably we can change the client level behavior.

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
stevie9868: For the rename, I will put up with a separate PR, thanks!

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -106,7 +115,13 @@ public TableDto create(final QualifiedName name, final TableDto tableDto) {
         log.info("Creating table {}", name);
         eventBus.post(new MetacatCreateTablePreEvent(name, metacatRequestContext, this, tableDto));
 
-        connectorTableServiceProxy.create(name, converterUtil.fromTableDto(tableDto));
+        final Optional<Runnable> unSaveOpOpt = saveParentChildRelationship(name, tableDto);
```
insyncoss, line 118: Nice refactor to make revert op a return value

insyncoss (approved):
Thanks for addressing comments, further considered changes can come in a later PR

metacat-common-server/src/main/java/com/netflix/metacat/common/server/usermetadata/ParentChildRelMetadataConstants.java
```diff
@@ -7,37 +7,46 @@
...
      * During get, the nested key specified in DefinitionMetadata[PARENTCHILDRELINFO] that indicates child infos.
      */
-    public static final String CHILDINFOS = "childInfos";
+    public static final String CHILD_INFOS = "childInfos";
+
+    /**
+     * During get, the nested key specified in DefinitionMetadata[PARENTCHILDRELINFO]
+     * that indicates if a table is parent.
+     */
+    public static final String IS_PARENT = "isParent";
```
amboz, line 49: yay

amboz (approved):
still +1 :]

PR #600, opened 2024-07-02
improve rename logic to use 2 phase commit idea

stevie9868:

PR #602, opened 2024-07-04
Yingjianw/additional safety rollout

stevie9868:
**First commit:**
1. Add a flag to controll the rollout in case there are any performance degradation or bug for each api call (create, rename, drop, get) that can touch the parent child relation service. This flag can also be used in the long run if we see performance degradation as we have more parent child relationship created and we can quickly block certain actions. 
2. Add a validation to make sure the parent name specified in the request actually exists in metacat
3. One bug fix on the rename side even though we are going to rewrite to use two phase commit idea in this pr https://github.com/Netflix/metacat/pull/600, which already handled this case. I have also added this case in our e2e test

More specifically the bug is:
- Supposed that we have childName = 'c' and parentName = 'p', and let's ignore uuid and clone type for example.
- Then we insert a new parent child relationship childName = 'nc' and parentName = 'p'
- So for now we have two records:
- 1. 'c' | 'p'
- 2. 'nc' | 'p'
- Then we rename 'c' to 'nc' because we always rename in parent child relationship first, and after the rename, the db will look something like
- 1. 'nc' | 'p'
- 2. 'nc' | 'p'
-  But it will fail later in the tableProxy as table 'nc' already and then if we revert back it will make all 2 records as 'c' | 'p'

**Second commit**
1. Add a validation during createTable to prevent creating a new record having the same name in the existing mysql db but uuid is different from that of the exiting one.

**Third commit**
1. Add a search endpoint for getParentsApi in case we need to block the get api and we can still have way to debug

**Forth commit**
1. Move out other parent child config out to keep all validation logic in the same place and also the config is general enough and doesn't need to be netflix specific

**Fifth commit**
Make all the flags to false by default

### DIFF
--- a/metacat-client/src/main/java/com/netflix/metacat/client/api/ParentChildRelV1.java
+++ b/metacat-client/src/main/java/com/netflix/metacat/client/api/ParentChildRelV1.java
@@ -7,7 +7,8 @@ import javax.ws.rs.GET;
 import javax.ws.rs.PathParam;
 
 import javax.ws.rs.core.MediaType;
-import com.netflix.metacat.common.dto.notifications.ChildInfoDto;
+import com.netflix.metacat.common.dto.ChildInfoDto;
+import com.netflix.metacat.common.dto.ParentInfoDto;
 
 import java.util.Set;
 
@@ -33,6 +34,26 @@ public interface ParentChildRelV1 {
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
     Set<ChildInfoDto> getChildren(
+        @PathParam("catalog-name")
+        String catalogName,
+        @PathParam("database-name")
+        String databaseName,
+        @PathParam("table-name")
+        String tableName
+    );
+
+    /**
+     * Return the list of parent.
+     * @param catalogName catalogName
+     * @param databaseName databaseName
+     * @param tableName tableName
+     * @return list of parent info dtos
+     */
+    @GET
+    @Path("parents/catalog/{catalog-name}/database/{database-name}/table/{table-name}")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    Set<ParentInfoDto> getParents(
         @PathParam("catalog-name")
         String catalogName,
         @PathParam("database-name")
--- a/metacat-main/src/main/java/com/netflix/metacat/main/api/v1/ParentChildRelController.java
+++ b/metacat-main/src/main/java/com/netflix/metacat/main/api/v1/ParentChildRelController.java
@@ -1,7 +1,8 @@
 package com.netflix.metacat.main.api.v1;
 
 import com.netflix.metacat.common.QualifiedName;
-import com.netflix.metacat.common.dto.notifications.ChildInfoDto;
+import com.netflix.metacat.common.dto.ChildInfoDto;
+import com.netflix.metacat.common.dto.ParentInfoDto;
 import com.netflix.metacat.common.server.usermetadata.ParentChildRelMetadataService;
 import com.netflix.metacat.main.api.RequestWrapper;
 import io.swagger.annotations.Api;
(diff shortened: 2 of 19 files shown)


metacat-client/src/main/java/com/netflix/metacat/client/api/ParentChildRelV1.java
```diff
@@ -40,4 +41,24 @@ Set<ChildInfoDto> getChildren(
...
+     * @param catalogName catalogName
+     * @param databaseName databaseName
+     * @param tableName tableName
+     * @return list of parent info dtos
+     */
+    @GET
+    @Path("parents/catalog/{catalog-name}/database/{database-name}/table/{table-name}")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    Set<ParentInfoDto> getParents(
```
amboz, line 56: one the one hand, i would suggest we rename this to `getParent` since we are enforcing the one-parent-per-child rule and its a little confusing as the name implies we can expect more than one results. on the other hand, this API does technically return more than one result, so we would need to have a different return type to make this change. i have a weak preference to rename it and change the contract despite it becoming less flexible, but its up to you

metacat-main/src/main/java/com/netflix/metacat/main/api/v1/ParentChildRelController.java
```diff
@@ -72,4 +73,35 @@ public Set<ChildInfoDto> getChildren(
             )
         );
     }
+
+    /**
+     * Return the list of children for a given table.
```
amboz, line 78:
```suggestion
     * Return the list of parent info for a given table.
```

metacat-main/src/main/java/com/netflix/metacat/main/api/v1/ParentChildRelController.java
```diff
@@ -72,4 +73,35 @@ public Set<ChildInfoDto> getChildren(
...
+     * @param catalogName catalogName
+     * @param databaseName databaseName
+     * @param tableName tableName
+     * @return list of childInfoDto
```
amboz, line 82:
```suggestion
     * @return list of parentInfoDto
```

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlParentChildRelMetaDataService.java
```diff
@@ -67,12 +88,77 @@ public MySqlParentChildRelMetaDataService(final JdbcTemplate jdbcTemplate, final
...
+                maxAllow = count;
+            }
+        }
+
+        // Finally fallback to the default value for all types
+        if (maxAllow == null) {
+            maxAllow = props.getMaxAllow();
+        }
+
+        // if maxAllow < 0, this means we can create as many child table under the parent
```
amboz, line 143:
```suggestion
        // if maxAllow < 0, this means we can create unlimited child tables under the parent
```

amboz (approved):

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -367,6 +376,17 @@ public TableDto deleteAndReturn(final QualifiedName name, final boolean isMView)
             }
         }
 
+        if (!config.isParentChildDropEnabled()) {
+            if (parentChildRelMetadataService.isChildTable(name)) {
+                throw new RuntimeException(name + " is a child table and "
+                    + "dropping a child table is currently disabled");
+            }
+            if (parentChildRelMetadataService.isParentTable(name)) {
```
insyncoss, line 384: Regardless of this check the dropping of parent table should be disabled anyways right

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlParentChildRelMetaDataService.java
```diff
@@ -67,12 +88,77 @@ public MySqlParentChildRelMetaDataService(final JdbcTemplate jdbcTemplate, final
...
+            }
+        }
+        return maxCloneAllow;
+    }
+
+    private void validateMaxAllow(final QualifiedName parentName,
+                                  final String type,
+                                  final ParentChildRelationshipProperties props) {
+        // Validate max clone allow
+        // First check if the parent table have configured max allowed on the table config
```
insyncoss, line 112:
```suggestion
        // First check if the parent table has a configured max allowed setting on the table config
```

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlParentChildRelMetaDataService.java
```diff
@@ -67,12 +88,77 @@ public MySqlParentChildRelMetaDataService(final JdbcTemplate jdbcTemplate, final
...
+        Integer maxAllow = getMaxAllowedFromNestedMap(
+            parentName,
+            type,
+            props.getMaxAllowPerTablePerRelType(),
+            true
+        );
+
+        // Then check if the parent have configured max allowed on the db config
```
insyncoss, line 120:
```suggestion
        // Then check if the parent has a configured max allowed setting on the db config
```

metacat-metadata-mysql/src/main/java/com/netflix/metacat/metadata/mysql/MySqlParentChildRelMetaDataService.java
```diff
@@ -67,12 +88,77 @@ public MySqlParentChildRelMetaDataService(final JdbcTemplate jdbcTemplate, final
...
+        if (maxAllow == null) {
+            maxAllow = getMaxAllowedFromNestedMap(
+                parentName,
+                type,
+                props.getMaxAllowPerDBPerRelType(),
+                false
+            );
+        }
+
+        // If not specified in maxAllowPerDBPerRelType,check the default max Allow based on relationType
```
insyncoss, line 130:
```suggestion
        // If not specified in maxAllowPerDBPerRelType, check the default maxAllow based on relationType
```

metacat-client/src/main/java/com/netflix/metacat/client/api/ParentChildRelV1.java
insyncoss: If the api stays having multiple return values then plural is better so that the correspondence holds, otherwise singular. There is no current use case for multiple parents but keeping the flexibility is reasonable.

insyncoss (approved):

stevie9868:
Thanks all, I will address the comments in the next PR.

PR #621, opened 2024-12-02
Ignore void partitions in metacat

gabeiglio:


amboz (approved):
can you copy over the pr description from what was reverted? ty!

PR #630, opened 2025-02-12
Upgrading to JDK17 and Spring Boot 3 and deprecating embedded hive client support

ursetta-netflix:


metacat-common/src/main/java/com/netflix/metacat/common/type/TypeSignature.java
```diff
@@ -119,17 +120,28 @@ public static TypeSignature parseTypeSignature(final String signature) {
             final char c = signature.charAt(i);
             if (c == '<') {
                 if (bracketCount == 0) {
-                    Preconditions.checkArgument(baseName == null, "Expected baseName to be null");
-                    Preconditions.checkArgument(parameterStart == -1, "Expected parameter start to be -1");
+                    if (baseName != null) {
```
insyncoss, line 123: Checking why does the Preconditions need to be removed for 1.7?

metacat-common/src/main/java/com/netflix/metacat/common/type/TypeSignature.java
```diff
@@ -19,7 +19,7 @@
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
-import com.google.common.base.Preconditions;
+//import com.google.common.base.Preconditions;
```
insyncoss, line 22: nit can be removed

metacat-common/src/main/java/com/netflix/metacat/common/type/TypeSignature.java
insyncoss: https://guava.dev/releases/17.0/api/docs/index.html?com/google/common/base/Preconditions.html

metacat-common/src/main/java/com/netflix/metacat/common/type/TypeRegistry.java
```diff
@@ -17,11 +17,11 @@
  */
 package com.netflix.metacat.common.type;
 
-import com.google.common.base.Preconditions;
```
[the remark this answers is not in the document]
ursetta-netflix: Guava versions are being resolved through SBN3 version preferences, it's pulling in a version where we are getting method not found on the Precondition Check

metacat-common/src/main/java/com/netflix/metacat/common/type/TypeSignature.java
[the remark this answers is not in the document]
ursetta-netflix: Guava versions are being resolved through SBN3 version preferences, it's pulling in a version where we are getting method not found on the Precondition Check

metacat-common/src/main/java/com/netflix/metacat/common/type/TypeSignature.java
ursetta-netflix: Guava versions are being resolved through SBN3 version preferences, it's pulling in a version where we are getting method not found on the Precondition Check

insyncoss (commented):
Thanks for walking through the changes. For preconditions what do you think about pinning the version instead of having it resolve through sbn dependencies?

insyncoss (approved):

PR #636, opened 2025-03-04
Adding common view support to the Polaris Connector

ldsantos0911:
- Added support for table parameters in Polaris
- Added support for handling common views during CREATE/READ/UPDATE in PolarisConnectorTableService
  - Repurposed the existing CommonViewHandler to replicate existing behavior.

### DIFF
--- a/metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
+++ b/metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
@@ -97,7 +97,7 @@ public class PolarisConnectorTableService implements ConnectorTableService {
         try {
             final PolarisTableEntity entity = polarisTableMapper.toEntity(tableInfo);
             polarisStoreService.createTable(entity.getDbName(), entity.getTblName(),
-                    entity.getMetadataLocation(), createdBy);
+                    entity.getMetadataLocation(), entity.getParams(), createdBy);
         } catch (DataIntegrityViolationException | InvalidMetaException exception) {
             throw new InvalidMetaException(name, exception);
         } catch (Exception exception) {
@@ -149,7 +149,9 @@ public class PolarisConnectorTableService implements ConnectorTableService {
                 .getTable(name.getDatabaseName(), name.getTableName())
                 .orElseThrow(() -> new TableNotFoundException(name));
             final TableInfo info = polarisTableMapper.toInfo(polarisTableEntity);
+            polarisTableMapper.decorateTableInfo(info);
             final String tableLoc = HiveTableUtil.getIcebergTableMetadataLocation(info);
+
             // Return the iceberg table with just the metadata location included if requested.
             if (connectorContext.getConfig().shouldFetchOnlyMetadataLocationEnabled()
                     && requestContext.isIncludeMetadataLocationOnly()) {
@@ -224,6 +226,12 @@ public class PolarisConnectorTableService implements ConnectorTableService {
                 log.warn("No parameters defined for iceberg table %s, no data update needed", name);
                 return;
             }
+            final Map<String, String> newTableParams = polarisTableMapper.filterMetadata(newTableMetadata);
+            final Map<String, String> existingTableParams = polarisStoreService
+                    .getTable(name.getDatabaseName(), name.getTableName())
+                    .orElseThrow(() -> new TableNotFoundException(name))
+                    .getParams();
+
             final String prevLoc = newTableMetadata.get(DirectSqlTable.PARAM_PREVIOUS_METADATA_LOCATION);
             final String newLoc = newTableMetadata.get(DirectSqlTable.PARAM_METADATA_LOCATION);
             if (StringUtils.isBlank(prevLoc)) {
@@ -246,12 +254,14 @@ public class PolarisConnectorTableService implements ConnectorTableService {
                 log.error(message);
                 throw new InvalidMetaException(name, message, null);
             }
-            // optimistically attempt to update metadata location
-            final boolean updated = polarisStoreService.updateTableMetadataLocation(
-                    name.getDatabaseName(), name.getTableName(),
-                    prevLoc, newLoc, lastModifiedBy);
+
+            // optimistically attempt to update metadata location and params
+            final boolean locationUpdated = polarisStoreService.updateTableMetadataLocationAndParams(
+                    name.getDatabaseName(), name.getTableName(), prevLoc, newLoc,
+                    existingTableParams, newTableParams, lastModifiedBy);
+
             // if succeeded then done, else try to figure out why and throw corresponding exception
-            if (updated) {
+            if (locationUpdated) {
                 requestContext.setIgnoreErrorsAfterUpdate(true);
                 log.warn("Success servicing Iceberg commit request for table: {}, "
                         + "previousLocation: {}, newLocation: {}",
@@ -348,7 +358,12 @@ public class PolarisConnectorTableService implements ConnectorTableService {
                 ConnectorUtils.sort(tbls, sort, Comparator.comparing(t -> t.getTblName()));
             }
             return ConnectorUtils.paginate(tbls, pageable).stream()
-                .map(t -> polarisTableMapper.toInfo(t)).collect(Collectors.toList());
+                .map(t -> {
+                        final TableInfo info = polarisTableMapper.toInfo(t);
+                        polarisTableMapper.decorateTableInfo(info);
+                        return info;
+                    }
+                ).collect(Collectors.toList());
         } catch (Exception exception) {
             final String msg = String.format("Failed polaris list tables %s using prefix %s", name, prefix);
             log.error(msg, exception);
@@ -360,7 +375,7 @@ public class PolarisConnectorTableService implements ConnectorTableService {
      * Return the table metadata from cache if exists else make the iceberg call and refresh it.
      * @param tableName             table name
      * @param tableMetadataLocation table metadata location
-     * @param info                  table info stored in hive metastore
+     * @param info                  table info stored in Polaris
      * @param includeInfoDetails    if true, will include more details like the manifest file content
      * @param useCache              true, if table can be retrieved from cache
      * @return TableInfo
--- a/metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
+++ b/metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
@@ -255,13 +255,13 @@ public class PolarisConnectorTableService implements ConnectorTableService {
                 throw new InvalidMetaException(name, message, null);
             }
 
-            // optimistically attempt to update metadata location and params
-            final boolean locationUpdated = polarisStoreService.updateTableMetadataLocationAndParams(
+            // optimistically attempt to update metadata location and/or params
+            final boolean updated = polarisStoreService.updateTableMetadataLocationAndParams(
                     name.getDatabaseName(), name.getTableName(), prevLoc, newLoc,
                     existingTableParams, newTableParams, lastModifiedBy);
 
             // if succeeded then done, else try to figure out why and throw corresponding exception
-            if (locationUpdated) {
+            if (updated) {
                 requestContext.setIgnoreErrorsAfterUpdate(true);
                 log.warn("Success servicing Iceberg commit request for table: {}, "
                         + "previousLocation: {}, newLocation: {}",
--- a/metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
+++ b/metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
@@ -18,7 +18,9 @@ import com.netflix.metacat.common.server.connectors.exception.TableNotFoundExcep
 import com.netflix.metacat.common.server.connectors.exception.TablePreconditionFailedException;
 import com.netflix.metacat.common.server.connectors.model.TableInfo;
 import com.netflix.metacat.common.server.properties.Config;
+import com.netflix.metacat.connector.hive.commonview.CommonViewHandler;
 import com.netflix.metacat.connector.hive.converters.HiveConnectorInfoConverter;
+import com.netflix.metacat.connector.hive.converters.HiveTypeConverter;
 import com.netflix.metacat.connector.hive.iceberg.IcebergTableHandler;
 import com.netflix.metacat.connector.hive.iceberg.IcebergTableWrapper;
 import com.netflix.metacat.connector.hive.sql.DirectSqlTable;
@@ -51,6 +53,7 @@ public class PolarisConnectorTableService implements ConnectorTableService {
     protected final HiveConnectorInfoConverter connectorConverter;
     protected final ConnectorContext connectorContext;
     protected final IcebergTableHandler icebergTableHandler;
+    protected final CommonViewHandler commonViewHandler;
     protected final PolarisTableMapper polarisTableMapper;
     protected final String catalogName;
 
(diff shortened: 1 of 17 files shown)


metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
```diff
@@ -348,7 +377,15 @@ public List<TableInfo> list(
                 ConnectorUtils.sort(tbls, sort, Comparator.comparing(t -> t.getTblName()));
             }
             return ConnectorUtils.paginate(tbls, pageable).stream()
-                .map(t -> polarisTableMapper.toInfo(t)).collect(Collectors.toList());
+                .map(t -> {
```
stevie9868, line 380: Do you think it's better to put decorate function inside the toInfo?

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
```diff
@@ -217,13 +233,24 @@ public void update(final ConnectorRequestContext requestContext, final TableInfo
...
+        } else {
+            icebergTableHandler.update(tableInfo);
+        }
         try {
             final Map<String, String> newTableMetadata = tableInfo.getMetadata();
             if (MapUtils.isEmpty(newTableMetadata)) {
                 log.warn("No parameters defined for iceberg table %s, no data update needed", name);
                 return;
             }
+            final Map<String, String> newTableParams = polarisTableMapper.filterMetadata(newTableMetadata);
```
stevie9868, line 248: Since we know the params are needed for only views now, do you think we can skip params for table for now?

One thing I am not sure is if we accidentally write params for table, we might see a behavior change.

metacat-functional-tests/src/functionalTest/groovy/com/netflix/metacat/MetacatSmokeSpec.groovy
```diff
@@ -108,6 +108,7 @@ class MetacatSmokeSpec extends Specification {
         def catalog = api.getCatalog(catalogName)
         if (!catalog.databases.contains(databaseName)) {
             api.createDatabase(catalogName, databaseName, new DatabaseCreateRequestDto())
+            Thread.sleep(5000)
```
stevie9868, line 111: sorry, this bad pattern was introduced by me. I will have a follow up work to not sleep during the test, thanks!

ldsantos0911: Apologies if there is a better way to go about it, I was just finding that the catalog didn't include the DB when I was debugging.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/mappers/PolarisTableMapper.java
```diff
@@ -34,33 +44,60 @@ public PolarisTableMapper(final String catalogName) {
...
+
     /**
      * {@inheritDoc}.
      */
     @Override
     public TableInfo toInfo(final PolarisTableEntity entity) {
         final int uriIndex = entity.getMetadataLocation().indexOf(PARAMETER_METADATA_PREFIX);
+
+        final HashMap<String, String> metadata = new HashMap<>(entity.getParams());
+        metadata.put(DirectSqlTable.PARAM_METADATA_LOCATION, entity.getMetadataLocation());
```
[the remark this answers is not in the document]
ldsantos0911: It is the pre-existing behavior and expected other places in the code. I've essentially moved the existing behavior which places the metadata location in TableInfo.metadata and adds some additional pieces of metadata out to a separate function.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
ldsantos0911: Decoration shouldn't be happening for views, so its a tradeoff between having a separate function for decoration or passing a flag for whether decoration should happen. I prefer having the `toInfo` be the pure conversion from Entity including all the metadata then we can remove/replace metadata as desired (i.e. via decoration), but I am open to changing it.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
ldsantos0911: This I'm not clear on. If we don't ever see ourselves wanting table params for use by metacat (before actually parsing the metadata) then maybe that could be fine? I don't have enough context at this moment to know if we would see issues from accepting other params for normal Iceberg tables. We wouldn't be returning them for get/list, but I don't know if there is some obscure code path which could change how metacat treats the tables. Furthermore, I don't know if there is an existing use pattern where a client is sending these params but expecting them not to be used.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/mappers/PolarisTableMapper.java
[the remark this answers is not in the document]
stevie9868: yeah, I see that.

I am wondering why PARAM_METADATA_LOCATION is kept around but others are removed but looks like you also move PARAM_METADATA_LOCATION to a separate function as well

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
stevie9868: yeah, I am also open to the current change.

The reason I raise this up is because I realize toInfo function and decorate function co-appears many different times so would like to centralize the logic into one place.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
stevie9868: yeah, my concern is more around potentially introducing new behavior as Metacat is a legacy system and unfortunately, many context are lost.

Client might put something new that is a non-reserved metacat property in the metadata, I believe for now we won't return the non-reserved metacat property in the metadata during get. But after this new change, we might now start returning those non reserved metacat properties introduced by the client.

tbh, I think the risk is relatively low so I am open to the changes.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/mappers/PolarisTableMapper.java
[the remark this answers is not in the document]
ldsantos0911: Ah, I got mixed up. PARAM_METADATA_LOCATION should always be there -- it's expected to be in the TableInfo metadata for views and tables alike. The other metadata that was being added here is being tacked on and isn't expected for views.  Since we don't currently support other parameters being returned for GET/LIST on tables, I added the `decorateTableInfo` function to replace the metadata (incl. params) which was retrieved from the DB with the expected metadata values we currently see. This includes the PARAM_METADATA_LOCATION, so in `decorateTableInfo` I'm adding it to the new metadata map so it is present after the metadata is overwritten.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
ldsantos0911: For GETs, the `PolarisTableMapper.decorateTableInfo` function replaces the TableInfo metadata with metadata that matches the current behavior, so we shouldn't see any unexpected params for GET/LIST. I believe that the new unit tests in PolarisConnectorTableServiceTest cover that particular case (maybe we could use a functional test too?). As for whether there is metacat code that might act differently given additional parameters, I don't think so, but I'm not positive.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
```diff
@@ -149,7 +155,13 @@ public TableInfo get(final ConnectorRequestContext requestContext, final Qualifi
                 .getTable(name.getDatabaseName(), name.getTableName())
                 .orElseThrow(() -> new TableNotFoundException(name));
             final TableInfo info = polarisTableMapper.toInfo(polarisTableEntity);
+            final boolean isView = connectorContext.getConfig().isCommonViewEnabled()
```
insyncoss, line 158: At this layer there shouldn't be logic to selectively enable views, it should always use view logic whenever the table is a view

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
```diff
@@ -217,13 +233,24 @@ public void update(final ConnectorRequestContext requestContext, final TableInfo
         final QualifiedName name = tableInfo.getName();
         final Config conf = connectorContext.getConfig();
         final String lastModifiedBy = PolarisUtils.getUserOrDefault(requestContext);
-        icebergTableHandler.update(tableInfo);
+        if (connectorContext.getConfig().isCommonViewEnabled()
```
insyncoss, line 236: At this layer there shouldn't be logic to selectively enable views, it should always use view logic whenever the table is a view

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
```diff
@@ -217,13 +233,24 @@ public void update(final ConnectorRequestContext requestContext, final TableInfo
...
+            icebergTableHandler.update(tableInfo);
+        }
         try {
             final Map<String, String> newTableMetadata = tableInfo.getMetadata();
             if (MapUtils.isEmpty(newTableMetadata)) {
                 log.warn("No parameters defined for iceberg table %s, no data update needed", name);
                 return;
             }
+            final Map<String, String> newTableParams = polarisTableMapper.filterMetadata(newTableMetadata);
+            final Map<String, String> existingTableParams = polarisStoreService
```
insyncoss, line 249: Let's optimize for the common case, and do not do this load if the table is not a view

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
```diff
@@ -246,10 +273,12 @@ public void update(final ConnectorRequestContext requestContext, final TableInfo
                 log.error(message);
                 throw new InvalidMetaException(name, message, null);
             }
-            // optimistically attempt to update metadata location
-            final boolean updated = polarisStoreService.updateTableMetadataLocation(
-                    name.getDatabaseName(), name.getTableName(),
-                    prevLoc, newLoc, lastModifiedBy);
+
+            // optimistically attempt to update metadata location and/or params
+            final boolean updated = polarisStoreService.updateTableMetadataLocationAndParams(
```
insyncoss, line 278: It would be better to have a branch for views and only update params if the table is a view to optimize common case, normal iceberg tables have no use for params currently.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/store/repos/PolarisTableRepository.java
```diff
@@ -91,4 +92,34 @@ int updateMetadataLocation(
...
+        + "WHERE t.metadataLocation = :expectedLocation "
+        + "AND ((t.params IS NULL AND :expectedParams IS NULL) OR t.params = :expectedParams)"
+        + "AND t.dbName = :dbName AND t.tblName = :tableName")
+    @Transactional
+    int updateMetadataLocationAndParams(
+        @Param("dbName") final String dbName,
+        @Param("tableName") final String tableName,
+        @Param("expectedLocation") final String expectedLocation,
+        @Param("newLocation") final String newLocation,
+        @Param("expectedParams") final Map<String, String> expectedParams,
```
insyncoss, line 121: There is no need for `expectedParams` because the analogy to `expectedLocation` doesn't really apply.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
ldsantos0911: Do you mean at the table connector level? I am mirroring the implementation in HiveConnectorFastTableService -- should we change it for Polaris?

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
ldsantos0911: Are we not intending to update params for all tables? I could be mistaken, but it looks like that is done currently in Hive. Do we want to restrict params to be used exclusively for views?

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
ldsantos0911: I think this answers my question above, but just to confirm we want to restrict params to only be used for views in Polaris?

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/store/repos/PolarisTableRepository.java
ldsantos0911: The current logic in HMS for table param updates locks the table param rows. Seeing as we want to perform only one query for Polaris updates as we had discussed offline, I thought that adding a comparison for the existing/expected `params` string would be a good extension of our existing compare-and-swap method for location updates. Would you recommend a different strategy?

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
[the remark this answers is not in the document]
ldsantos0911: Removed this flag from the isView check -- resolving thread.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
ldsantos0911: Refactored a bit and no longer need decorate function. Resolving thread.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
ldsantos0911: Per offline discussions, we won't be using table params for tables, just views. Most recent commit implements that. Resolving thread.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
ldsantos0911: Done in new commit.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/store/repos/PolarisTableRepository.java
ldsantos0911: Discussed offline -- won't do a comparison for the params, although we will update in the same statement. Resolving thread.

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
ldsantos0911: Removed the config check. Resolving thread.

stevie9868 (approved):

insyncoss:
Thanks for making the changes to separate out view based code

insyncoss (approved):

PR #645, opened 2025-04-17
Yingjianw/vtts validation on sql level

stevie9868:


metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -684,11 +684,6 @@ public TableDto updateAndReturn(final QualifiedName name, final TableDto tableDt
                 log.info("Saving user metadata for table {}", name);
                 final long start = registry.clock().wallTime();
                 userMetadataService.saveMetadata(metacatRequestContext.getUserName(), tableDto, true);
-                final long vtts = MetacatUtils.getVtts(tableDto.getDefinitionMetadata());
```
stevie9868, line 687: This was a previous hack to reduce the probability of vtts setting backward but it is not effective and does not solve the root issue so remove this code

PR #649, opened 2025-04-23
select definitionMetada with update

stevie9868:
This PR adds 2 new implementations:
1. Select for update for saveDefinitionMetadata as the current implementation can lead to lost of update. Using select For Update will hold a write lock which prevents other transactions concurrently read and then update the value. For simple select, this won't block for Isolation level that is weaker than serializable.
3. Add a preMergeInterceptor which happens before merging the definitionMetadata so we can pre-process the user definitionMetadata blob.

### DIFF
--- a/metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
+++ b/metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
@@ -684,11 +684,6 @@ public class TableServiceImpl implements TableService {
                 log.info("Saving user metadata for table {}", name);
                 final long start = registry.clock().wallTime();
                 userMetadataService.saveMetadata(metacatRequestContext.getUserName(), tableDto, true);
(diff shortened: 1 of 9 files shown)


metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -684,11 +684,6 @@ public TableDto updateAndReturn(final QualifiedName name, final TableDto tableDt
                 log.info("Saving user metadata for table {}", name);
                 final long start = registry.clock().wallTime();
                 userMetadataService.saveMetadata(metacatRequestContext.getUserName(), tableDto, true);
-                final long vtts = MetacatUtils.getVtts(tableDto.getDefinitionMetadata());
```
stevie9868, line 687: get rid of this previous hack as it does not work well in practice

ldsantos0911 (approved):

ursetta-netflix (approved):

PR #665, opened 2025-08-25
[BDW-887] Block Writes for Incompatible Iceberg Clients

gtrettenero:
Block updates for incompatible iceberg clients (Iceberg clients with versions lower than 1.1) if the client is trying to update a table with branches or tags. Older client can unintentionally override branches and tags since they're not aware of them, so only newer clients can update tables that are using these features.

### DIFF
--- a/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableWrapper.java
+++ b/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableWrapper.java
@@ -19,6 +19,8 @@ package com.netflix.metacat.connector.hive.iceberg;
 import org.apache.iceberg.Table;
 import lombok.Data;
 
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 
@@ -53,20 +55,15 @@ public class IcebergTableWrapper {
     }
 
     /**
-     * Check if the table has any non-main branches or tags.
-     * @return true if the table has non-main branches or tags
-     */
-    public boolean hasNonMainBranchesOrTags() {
-        return hasNonMainBranches() || hasTags();
-    }
-
-    /**
-     * Populate branch/tag metadata into extraProperties.
+     * Populate and return branch/tag metadata properties.
      * This should be called explicitly when metadata injection is needed.
+     * @return map containing branch/tag metadata properties
      */
-    public void populateBranchTagMetadata() {
-        this.extraProperties.put(ICEBERG_HAS_NON_MAIN_BRANCHES_KEY, String.valueOf(hasNonMainBranches()));
-        this.extraProperties.put(ICEBERG_HAS_TAGS_KEY, String.valueOf(hasTags()));
+    public Map<String, String> populateBranchTagMetadata() {
+        final Map<String, String> branchTagMetadata = new HashMap<>();
+        branchTagMetadata.put(ICEBERG_HAS_NON_MAIN_BRANCHES_KEY, String.valueOf(hasNonMainBranches()));
+        branchTagMetadata.put(ICEBERG_HAS_TAGS_KEY, String.valueOf(hasTags()));
+        return branchTagMetadata;
     }
 
     /**
@@ -94,10 +91,13 @@ public class IcebergTableWrapper {
      * @throws RuntimeException if unable to read table references
      */
     private Set<String> extractBranches() {
-            final var refs = table.refs();
-            return refs.keySet().stream()
-                .filter(ref -> refs.get(ref).isBranch())
-                .collect(java.util.stream.Collectors.toSet());
+        final var refs = table.refs();
+        if (refs == null || refs.isEmpty()) {
+            return Collections.emptySet();
+        }
+        return refs.keySet().stream()
+            .filter(ref -> refs.get(ref).isBranch())
+            .collect(java.util.stream.Collectors.toSet());
     }
 
     /**
@@ -106,9 +106,12 @@ public class IcebergTableWrapper {
      * @throws RuntimeException if unable to read table references
      */
     private Set<String> extractTags() {
-            final var refs = table.refs();
-            return refs.keySet().stream()
-                .filter(ref -> refs.get(ref).isTag())
-                .collect(java.util.stream.Collectors.toSet());
+        final var refs = table.refs();
+            if (refs == null || refs.isEmpty()) {
+                return Collections.emptySet();
+            }
+        return refs.keySet().stream()
+            .filter(ref -> refs.get(ref).isTag())
+            .collect(java.util.stream.Collectors.toSet());
     }
 }
--- a/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableWrapper.java
+++ b/metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableWrapper.java
@@ -20,12 +20,95 @@ import org.apache.iceberg.Table;
 import lombok.Data;
 
 import java.util.Map;
+import java.util.Set;
 
 /**
  * This class represents the iceberg table.
  */
 @Data
 public class IcebergTableWrapper {
+    /** Key for indicating if the table has non-main branches. */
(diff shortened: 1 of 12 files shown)


metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableWrapper.java
```diff
@@ -17,15 +17,103 @@
...
+        this.branches = extractBranches();
+        this.tags = extractTags();
+    }
+
+    /**
+     * Check if the table has any branches (excluding the default main branch).
+     * @return true if the table has branches other than main
+     */
+    public boolean hasBranches() {
+        return !branches.isEmpty() && !(branches.size() == 1 && branches.contains("main"));
```
gabeiglio, line 52: This is a bit confusing imo. I think `!branches.isEmpty()`  is redundant? Since in any other case that branches > 1 then we now we have branches other than main?

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
```diff
@@ -236,6 +236,10 @@ public List<QualifiedName> listNames(
     @Override
     public void update(final ConnectorRequestContext requestContext, final TableInfo tableInfo) {
         final QualifiedName name = tableInfo.getName();
+
+        // Validate Iceberg branches/tags support for client compatibility
+        icebergTableHandler.validateIcebergBranchesTagsSupport(requestContext, name, tableInfo);
```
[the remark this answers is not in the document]
yingjianwu98: This [call](https://github.com/Netflix/metacat/blob/3ef4904e48fbad996a8126b368ce4ac58f4f4f83/metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java#L873) is already calling the icebergHandler so no need to call it again.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/converter/ConverterUtil.java
```diff
@@ -231,7 +231,12 @@ public PartitionInfo fromPartitionDto(final PartitionDto partitionDto) {
      * @return connector context
      */
     public ConnectorRequestContext toConnectorContext(final MetacatRequestContext metacatRequestContext) {
-        return mapper.map(metacatRequestContext, ConnectorRequestContext.class);
+        final ConnectorRequestContext context = mapper.map(metacatRequestContext, ConnectorRequestContext.class);
```
amboz, line 234: assuming we move the validation to the private repo this is not needed

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableHandler.java
```diff
@@ -438,4 +439,170 @@ private String getIcebergPartitionURI(final String databaseName,
...
+     * Validates that the client supports Iceberg branches and tags if the table has them.
+     * This prevents older clients (e.g., Iceberg 0.9) from modifying tables that contain
+     * branches or tags, which could result in data loss.
+     *
+     * @param requestContext the connector request context containing headers
+     * @param name the qualified table name
+     * @param tableInfo the table info
+     * @throws UnsupportedClientOperationException if the client doesn't support branches/tags but the table has them
+     */
+    public void validateIcebergBranchesTagsSupport(final ConnectorRequestContext requestContext,
```
amboz, line 453: we should attempt to do this in the private repo

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisConnectorTableService.java
```diff
@@ -236,6 +236,10 @@ public List<QualifiedName> listNames(
     @Override
     public void update(final ConnectorRequestContext requestContext, final TableInfo tableInfo) {
         final QualifiedName name = tableInfo.getName();
+
+        // Validate Iceberg branches/tags support for client compatibility
```
amboz, line 240: this may be a moot point if we move this validation out of OSS, but we should add a test that checks whether the table service will actually swallow the unsupported client exception in certain cases and not propagate to the user (undesired behavior)

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableWrapper.java
```diff
@@ -17,15 +17,108 @@
...
+     */
+    public boolean hasBranchesOrTags() {
+        return hasBranches() || hasTags();
+    }
+
+    /**
+     * Get summary information about branches and tags for logging/debugging.
+     * @return formatted string with branch and tag counts and names
+     */
+    public String getBranchesAndTagsSummary() {
```
yingjianwu98, line 78: hmm, wondering where are we going to use this?

gtrettenero: this was helpful for debugging and I was thinking for future we can add it to error logs if we start seeing issues with branches and/or tags

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableWrapper.java
```diff
@@ -17,15 +17,108 @@
...
+import lombok.Getter;
 
 import java.util.Map;
+import java.util.Set;
 
 /**
  * This class represents the iceberg table.
  */
-@Data
+@Getter
```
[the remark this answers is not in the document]
yingjianwu98: ah you could add the @Builder so you can build up the object even if the variable is final,

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableWrapper.java
```diff
@@ -20,12 +20,95 @@
...
+    }
+
+    /**
+     * Extract branch names from the table references.
+     * @return set of branch names
+     * @throws RuntimeException if unable to read table references
+     */
+    private Set<String> extractBranches() {
+            final var refs = table.refs();
+            return refs.keySet().stream()
```
yingjianwu98, line 98: After some consideration, I feel like let's add the check on if refs is equal to null, we just return an Empty Set. For other exceptions, we should fail.

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableHandler.java
```diff
@@ -438,4 +439,170 @@ private String getIcebergPartitionURI(final String databaseName,
...
+            return;
+        }
+
+        final Map<String, String> headers = requestContext.getAdditionalContext();
+        if (headers == null) {
+            blockUnsupportedClient(name, "No headers provided");
+            return;
+        }
+
+        // Check for Iceberg REST catalog spec version (0.14.1+ supports branches/tags)
```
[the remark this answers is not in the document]
gtrettenero: Ah this is for cases when IRC communicates with Metacat since IRC won't have the Netflix Client Version that we have. I changed this to internal though so not relevant for this PR anymore

metacat-connector-hive/src/main/java/com/netflix/metacat/connector/hive/iceberg/IcebergTableWrapper.java
```diff
@@ -20,12 +20,95 @@
...
+    public boolean hasTags() {
+        final Set<String> tags = extractTags();
+        return !tags.isEmpty();
+    }
+
+    /**
+     * Check if the table has any non-main branches or tags.
+     * @return true if the table has non-main branches or tags
+     */
+    public boolean hasNonMainBranchesOrTags() {
```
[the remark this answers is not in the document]
gtrettenero: not explicitly anymore since we split up the tags thought I could keep it in case it was useful alter but I can remove it for now since it's easy to reimplement if needed

PR #667, opened 2025-08-28
exclude table table from request tags

yingjianwu98:
Having tableName level is not useful and could cause delay of metrics in our system


ldsantos0911 (approved):
Small nit, otherwise LGTM!

PR #676, opened 2025-10-07
Adding support for DBS_PARAMS in the polaris connector so it is in parity with hms

ursetta-netflix:
Adding support for key/value metadata on polaris dbs in parity of older HMS dbs. This will support full polaris migration and allow us to support the existing HMS dbs behavior.

- Adding new `DBS_PARAMS` schema
- Adding params `Map<String, String>` to polaris DBS entity to support key value metadata
- Updating Mappers/Connectors/Store to read/write params when present

Note, this now allows metadata to be written to polaris dbs but does not accept any other updates. This matches the existing behavior where metadata provided during update will overwrite all key/values. If metadata is not provided it is a No-Op against existing key/values.


metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/mappers/PolarisDatabaseMapper.java
```diff
@@ -31,6 +33,7 @@ public DatabaseInfo toInfo(final PolarisDatabaseEntity entity) {
         final DatabaseInfo databaseInfo = DatabaseInfo.builder()
             .name(QualifiedName.ofDatabase(catalogName, entity.getDbName()))
             .auditInfo(mapper.toInfo(entity.getAudit()))
+            .metadata(entity.getParams() != null ? entity.getParams() : new HashMap<>())
```
ldsantos0911, line 36: nit: I don't think it should be necessary to have these checks to convert `null`s to empty `HashMap`s. `StringParamsConverter` converts `null` db values to `new HashMap<>()` entity values and `null` or empty entity values to `null` db values.

ldsantos0911 (approved):
Minor nit (take it or leave it), but LGTM!

PR #693, opened 2026-03-30
Add retry with exponential backoff to thrift server socket binding

ursetta-netflix:
Thrift server startup intermittently fails (~5%) on fresh container startup with TTransportException when binding to the configured port. This is likely caused by a race with container networking or port TIME_WAIT from IP reuse.

Wrap TServerSocket creation in a Guava Retryer (5 attempts, exponential backoff capped at 30s) to handle transient bind failures gracefully.

### DIFF
--- a/metacat-thrift/build.gradle
+++ b/metacat-thrift/build.gradle
@@ -33,6 +33,7 @@ dependencies {
      *******************************/
 
     api("com.google.guava:guava")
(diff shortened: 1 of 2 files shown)


gtrettenero (approved):
LGTM, if we still see some issues even with the retryer we might want to dive into the race condition you mentioned to avoid the transient error in the first place

PR #700, opened 2026-04-28
Support compound app:userName rules in connector authorization

ursetta-netflix:
## Summary                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                        
  Extends connector authorization (`connector.authorized-callers`) to support compound rules of the form `app:userName`, in addition to the existing plain `app` entries.                                                                                               
                                      
  - **Plain entry** (e.g. `app1`) — matches on `SsoDirectCallerAppName` only. Unchanged behavior.                                                                                                                                                                        
  - **Compound entry** (e.g. `app2:user2`) — requires **both** `SsoDirectCallerAppName` AND the request `userName` to match exactly.                                                                                       
                                                                                                                                                                                                                                                                        
  ## Motivation                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                        
  After #690, connector authorization checks `SsoDirectCallerAppName` against a flat allow list. This is too coarse for callers like:                                                                                                                                   
                                                                                                                                                                                                                                  
  MetacatRequestContext{userName='user2', ..., additionalContext='{SsoDirectCallerAppName=app2, ...}'}                                                                                                                                                           
                                                                                                                                                                                                                                  
  Allowing all of `app2` would over-grant. We want a way to grant access only when both the direct caller app and the userName line up — e.g. `app2` + `user2` only — while leaving existing app-only entries working for other callers.                       
                                                                                                                                                                                                                                                                        
  ## Changes                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                        
  - **New** `AuthorizedCaller` value class (`appName` required, `userName` nullable) with a `matches(callerApp, callerUserName)` method.                                                                                                                                
  - `ConnectorFactoryDecorator` now parses each comma token, splitting on the first `:` to build compound rules. Empty tokens are skipped; matching remains case-sensitive.                                                       
  - `ConnectorAuthorizationUtil.checkAuthorization` reads `userName` from the request context and uses `anyMatch` over the allow-list rules. Deny logs include the userName for easier debugging.                                                                       
  - The four `Authorizing*Service` decorators (`Catalog`, `Database`, `Table`, `Partition`) now hold `Set<AuthorizedCaller>` instead of `Set<String>`.                                                                                                                  
                                                                                                                                                                                                                                                                        
  ## Config example                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                        
  connector.authorization-required=true                                                                                                                                                                                                                                 
  connector.authorized-callers=app1,app2,app3:user1                                                                                                                              
                                                                                                                                                                                                                                                                        
  Behavior with the above:                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                        
  | `SsoDirectCallerAppName` | `userName`   | Result   |                                                                                                                                                                                                                
  | ------------------------ | ------------ | -------- |                                                                                                                                                                                                                
  | `app1`                    | any          | allowed  |                                                                                                                                                                                                                
  | `app3`                 | `user1` | allowed  |                                                                                                                                                                          
  | `app3`                 | `user2`      | denied   |                                                                                                                                                                                                                
  | `app3`                 | `null`       | denied   |                                                                                                                                                                          
  | `other`                  | `user2` | denied   |                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                        
  ## Backwards compatibility                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                        
  Configs with no `:` in any entry parse to rules with `userName=null` and behave identically to the previous implementation. No deployment-time action needed for existing connectors.                                                                                 
                                                                                                                                                                                                                                  
  ## Test plan                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                  
  - [x] `./gradlew :metacat-common-server:test` — all targeted specs pass                                                                                                                                                                                               
  - [x] New cases in `ConnectorAuthorizationUtilSpec`: compound match, userName mismatch, null userName, mixed plain+compound list, app mismatch
  - [x] New parsing test in `ConnectorFactoryDecoratorSpec` for `app:userName` syntax                                                                                                                                                                                   
  - [x] Existing `Authorizing*ServiceSpec` tests migrated to `AuthorizedCaller` and continue to pass

### DIFF
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorFactoryDecorator.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorFactoryDecorator.java
@@ -18,6 +18,7 @@
 package com.netflix.metacat.common.server.connectors;
 
 import com.netflix.metacat.common.server.api.ratelimiter.RateLimiter;
+import com.netflix.metacat.common.server.util.AuthorizedCaller;
 import lombok.Getter;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
@@ -55,7 +56,7 @@ public class ConnectorFactoryDecorator implements ConnectorFactory {
     private final RateLimiter rateLimiter;
     private final boolean rateLimiterEnabled;
     private final boolean authorizationRequired;
-    private final Set<String> authorizedCallers;
+    private final Set<AuthorizedCaller> authorizedCallers;
 
     /**
      * Creates the decorated connector factory that wraps connector services
@@ -225,7 +226,7 @@ public class ConnectorFactoryDecorator implements ConnectorFactory {
         );
     }
 
-    private Set<String> getAuthorizedCallers() {
+    private Set<AuthorizedCaller> getAuthorizedCallers() {
         if (connectorContext.getConfiguration() == null) {
             return Collections.emptySet();
         }
@@ -237,11 +238,24 @@ public class ConnectorFactoryDecorator implements ConnectorFactory {
             return Collections.emptySet();
         }
 
-        // Split by comma, trim whitespace, filter empty strings, collect to set
+        // Split by comma, trim whitespace, filter empty strings, parse each token.
+        // Plain entry "app" matches on SsoDirectCallerAppName only.
+        // Compound entry "app:userName" requires both SsoDirectCallerAppName AND userName to match.
         // Note: matching is case-sensitive (e.g., "IRC" will not match "irc")
         return Arrays.stream(callers.split(","))
             .map(String::trim)
             .filter(s -> !s.isEmpty())
+            .map(ConnectorFactoryDecorator::parseAuthorizedCaller)
             .collect(Collectors.toSet());
+    }
+
+    private static AuthorizedCaller parseAuthorizedCaller(final String token) {
+        final int colon = token.indexOf(':');
+        if (colon < 0) {
+            return new AuthorizedCaller(token, null);
+        }
+        final String app = token.substring(0, colon);
+        final String user = token.substring(colon + 1);
+        return new AuthorizedCaller(app, user);
     }
 }
--- a/metacat-common-server/src/main/java/com/netflix/metacat/common/server/util/AuthorizedCaller.java
+++ b/metacat-common-server/src/main/java/com/netflix/metacat/common/server/util/AuthorizedCaller.java
@@ -1,0 +1,64 @@
+/*
+ *
+ *  Copyright 2024 Netflix, Inc.
+ *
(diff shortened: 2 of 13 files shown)


metacat-common-server/src/main/java/com/netflix/metacat/common/server/util/AuthorizedCaller.java
```diff
@@ -0,0 +1,64 @@
...
+     *
+     * @param callerAppName  the SsoDirectCallerAppName from the request
+     * @param callerUserName the userName from the request (may be null)
+     * @return true if the caller is allowed by this rule
+     */
+    public boolean matches(final String callerAppName, @Nullable final String callerUserName) {
+        if (!appName.equals(callerAppName)) {
+            return false;
+        }
+        if (userName == null) {
```
gtrettenero, line 59: are there any edge cases where the username filed is not a null value but it's populated with a value that should be treated as null?

gtrettenero: example having "null" as user for example

FrancisGodinho: follow up, should an empty string be considered as null here?

ursetta-netflix: username here is from the catalog configuration like `user` in `app:user`. So this will always either be a value we set, or null if it's not colon separated like `app2`.

So we will only compare a string we set ie: `user` to the `callerUserName` which is in the request. So we are expecting the request to explicitly match the known string or we deny.

gtrettenero (approved):

metacat-common-server/src/main/java/com/netflix/metacat/common/server/connectors/ConnectorFactoryDecorator.java
```diff
@@ -237,11 +238,24 @@ private Set<String> getAuthorizedCallers() {
...
             .collect(Collectors.toSet());
     }
+
+    private static AuthorizedCaller parseAuthorizedCaller(final String token) {
+        final int colon = token.indexOf(':');
+        if (colon < 0) {
+            return new AuthorizedCaller(token, null);
+        }
+        final String app = token.substring(0, colon);
+        final String user = token.substring(colon + 1);
```
gabeiglio, line 258: wondering if a token like `someapp:` is possible?

ursetta-netflix: We won't ever set it that way. But I can add an additional rule to deny
that explicitly so it's not possible

PR #701, opened 2026-04-30
increasing thrift connection wait time

FrancisGodinho:
Increasing the thrift server connection wait time to decrease connection failure on startup


ursetta-netflix (approved):

gtrettenero:
if we're still seeing the error is there a need to revert the old change or nah?

FrancisGodinho:
> if we're still seeing the error is there a need to revert the old change or nah?

i think retrying connection isn't a bad thing to have regardless since there can always be some intermittent connection issues (even if that isn't the cause of the consistent startup failure).  We may want to revert the wait time slightly so that its not as long though.

PR #704, opened 2026-06-04
Skip metadata for sv load

ldsantos0911:
Loading a secure view via Metacat will fail currently because Metacat cannot parse the OSS view format. This forces Polaris to treat secure view loads as though includeMetadataLocationOnly is enabled. We also disable secure view creation. Will quickly verify during deployment.


yingjianwu98:
btw, we support deployment to preview environment now.

if you only need dev db, start your pr with feature/
if you need production db, start your pr with feature_proddb/

yingjianwu98 (approved):

ldsantos0911:
> btw, we support deployment to preview environment now.
> 
> if you only need dev db, start your pr with feature/ if you need production db, start your pr with feature_proddb/

Oh cool, will open a different PR

PR #707, opened 2026-07-09
Emit Iceberg current_snapshot_id on the table DTO

gtrettenero:
## What

Surface the Iceberg **current snapshot id** onto the `TableDto` (metadata key `current_snapshot_id`) whenever an Iceberg table is read, and exclude it from persistence as a derived, read-only value.

## Why

Downstream consumers need to tell a **metadata-only** table update (schema/property/tag change) apart from a **data** commit — e.g. to avoid triggering compaction/rewrite/CDC for changes that wrote no data. Iceberg bumps `metadata_location` on *every* commit, so location alone can't distinguish them; the **current snapshot id** only advances on a data commit and is the correct signal.

### Design note (changed from an earlier revision)

An earlier version of this branch had metacat itself derive a `metadataOnly` boolean and emit it on the SNS/AP `TABLE_UPDATE` payload. That was dropped in favor of emitting just the raw `current_snapshot_id`, because:

- It **piggybacks existing plumbing**: the AP-events producer serializes the whole `TableDto` (`previous_table_json`/`current_table_json` are opaque JSON), and the fan-out only *removes* one key — so a new metadata-map key propagates to consumers with **no proto or fan-out change**. A first-class `metadataOnly` field would instead require coordinated edits across the proto + producer + both fan-out re-mappers.
- It **keeps the policy where the data lives**: the consumer decides what "metadata-only" means (and can evolve that logic) without a metacat redeploy.

## How

- `HiveConnectorInfoConverter.fromIcebergTableToTableInfo` writes `current_snapshot_id` (value `-1` when there is no current snapshot) — this is the shared converter, so it applies on both the Hive and Polaris read paths.
- `DirectSqlTable.PARAM_CURRENT_SNAPSHOT_ID` and `PolarisTableMapper` exclude it from persistence (derived on read only).
- `TableDto.CURRENT_SNAPSHOT_ID_METADATA_KEY` + `TableDto.getCurrentSnapshotId()` for typed access (empty for non-Iceberg tables/views).

## Test

- Unit: `TableDtoSpec` (snapshot-id parsing), `HiveConnectorInfoConvertorSpec` (converter surfacing).
- Functional: `MetacatSmokeSpec` end-to-end test that `current_snapshot_id` is emitted and transitions correctly across the table lifecycle (empty → data write → metadata-only commit → data write) on both the Hive and Polaris catalogs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatUpdateTablePostEvent.java
```diff
@@ -82,4 +83,29 @@ public MetacatUpdateTablePostEvent(
...
+     *
+     * @return true if the main-branch current snapshot id is unchanged; false otherwise or if unknown
+     */
+    public boolean getMetadataOnlyUpdate() {
+        if (!isLatestCurrentTable) {
+            return false;
+        }
+        final Optional<Long> oldSnapshotId = oldTable.getCurrentSnapshotId();
+        final Optional<Long> currentSnapshotId = currentTable.getCurrentSnapshotId();
+        if (oldSnapshotId.isEmpty() || currentSnapshotId.isEmpty()) {
```
yingjianwu98: if both are empty, shouldn't this return true?

Also what about oldSnapshot is not Empty and new is empty? This is kind of illegal state but I guess we can return false to be safe

yingjianwu98:
Also make sure to double check the backward compatible with the event schema.

metacat-common-server/src/main/java/com/netflix/metacat/common/server/events/MetacatUpdateTablePostEvent.java
gtrettenero: ah well for iceberg tables we populate an 'empty' snapshot as -1. so this would technically be empty only for cases like with non-iceberg tables or views. But in any case it's a good point that if both are empty they should be true so I'll change this to just a straight comparison between the ids

FrancisGodinho:
did we verify that MetacatJson serialization still works? (i.e. go from TableDto -> JSON and then from JSON->TableDto) and make sure that snapshot_id is preserved throughout?

gtrettenero:
@FrancisGodinho added a test [here](https://github.com/Netflix/metacat/pull/707/changes#diff-358db5ef8eb55c0b76b565a258dac3221a9adb61614ba63d728251d22464580dR52)

yingjianwu98 (approved):

metacat-functional-tests/src/functionalTest/groovy/com/netflix/metacat/MetacatSmokeSpec.groovy
```diff
@@ -765,6 +765,103 @@ class MetacatSmokeSpec extends Specification {
         catalogName << ['polaris-metastore', 'hive-metastore']
     }
 
+    @Unroll
+    def "Test current_snapshot_id is emitted end-to-end across the iceberg lifecycle in #catalogName"() {
```
bryanck: It seems like we could just add the snapshot ID check to an existing test, we don't need a whole separate test for a single field?

gtrettenero: Good call — dropped the standalone tests and folded the snapshot-id assertions into the existing "Test create/update iceberg table" test (verifying it's emitted and transitions across the two data commits). Also trimmed the verbose comment blocks across the change.

PR #710, opened 2026-08-06
fix inefficient query

yingjianwu98:
1. Listing from limit/offset to keyset-based pagination to prevent O(N^2) look up.
2. Some minor clean up on unnecessary classes since crdb is removed.
3. Since, polaris always return the ascending order name so simplify the existing logic.


metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/store/jdbc/PolarisDatabaseReplicaJDBC.java
```diff
@@ -1,96 +1,108 @@
...
-                            .dbName(rs.getString("name"))
-                            .location(rs.getString("location"))
-                            .audit(audit)
-                            .build();
-                    } else {
-                        return (T) rs.getString("name");
-                    }
-                }
-            });
+        final String prefix = (dbNamePrefix == null ? "" : dbNamePrefix) + "%";
```
gtrettenero, line 88: why is it that we allow null?

yingjianwu98: It's a legacy behavior....if you don't pass a prefix, it means you want all the db...

gtrettenero (approved):

PR #712, opened 2026-08-10
feature: Implement alias db lookups

gabeiglio:


metacat-connector-polaris/src/functionalTest/scripts/datastores/aurora/docker-entrypoint-initdb-primary.d/schema.sql
```diff
@@ -37,3 +38,15 @@ CREATE TABLE TBLS (
                     CONSTRAINT unique_catalog_db_tbl UNIQUE (catalog_name, db_name, tbl_name),
                     CONSTRAINT fk_tbls_db FOREIGN KEY (catalog_name, db_name) REFERENCES DBS(catalog_name, name) ON DELETE RESTRICT ON UPDATE CASCADE
 );
+
+CREATE TABLE ALIASES (
+                    id UUID DEFAULT uuid_generate_v4() NOT NULL PRIMARY KEY,
```
bryanck, line 43: Seems like we aren't consistent in using uuid_generate_v4 vs gen_random_uuid, but that's preexisting

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisAliasService.java
```diff
@@ -0,0 +1,72 @@
...
+     *
+     * @param aliasRepository the alias repository
+     * @param ttlSeconds      how long a resolution is cached for, in seconds
+     * @param maxSize         maximum number of names held in the cache
+     */
+    public PolarisAliasService(final PolarisAliasRepository aliasRepository,
+                               final long ttlSeconds,
+                               final long maxSize) {
+        this.aliasRepository = aliasRepository;
+        this.cache = CacheBuilder.newBuilder()
```
bryanck: I don't think the cache is very useful unless we make it a shared cache, but I suppose it doesn't hurt

metacat-functional-tests/metacat-test-cluster/docker-compose-aurora.yml
```diff
@@ -70,6 +70,7 @@ services:
                 -Dmetacat.definition.metadata.delete.enableDeleteForQualifiedNames=hive-metastore/hsmoke_ddb,hive-metastore/hsmoke_ddb1/test_create_table1,embedded-hive-metastore,embedded-fast-hive-metastore/fsmoke_db1,embedded-fast-hive-metastore/fsmoke_ddb1,embedded-fast-hive-metastore/shard,embedded-fast-hive-metastore/fsmoke_db4,s3-mysql-db,mysql-56-db
                 -Dmetacat.hive.metastore.batchSize=10
                 -Dmetacat.hive.iceberg.enabled=true
+                -Dmetacat.alias-service-properties.enabled=true
```
bryanck, line 73: maybe just `aliases` instead of `alias-service-properties`

metacat-connector-polaris/src/main/java/com/netflix/metacat/connector/polaris/PolarisAliasService.java
bryanck: Also, should this respect `metacat.cache.enabled`?

gabeiglio: Ill probably just delete the caching logic to make the change a bit simpler

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -111,6 +114,11 @@ public class TableServiceImpl implements TableService {
     public TableDto create(final QualifiedName name, final TableDto tableDto) {
         final MetacatRequestContext metacatRequestContext = MetacatContextManager.getContext();
         validate(name);
+
+        if (aliasService.isAlias(name)) {
```
bryanck: `aliasService.isAlias()` performs a lookup, perhaps if aliases are disabled that lookup should be skipped

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
```diff
@@ -875,9 +903,15 @@ public List<QualifiedName> getQualifiedNames(final QualifiedName name,
     }
 
     @Override
-    public TableDto updateAndReturn(final QualifiedName name, final TableDto tableDto,
+    public TableDto updateAndReturn(final QualifiedName requestedName, final TableDto tableDto,
                                     final boolean shouldThrowExceptionOnMetadataSaveFailure) {
+        final QualifiedName name = resolveAlias(requestedName);
         validate(name);
+
+        Preconditions.checkArgument(name.getTableName().equalsIgnoreCase(tableDto.getName().getTableName()),
```
bryanck: This only compares the table name and not catalog or namespace

bryanck: Also slight change of behavior in that a `null` DTO name will cause an NPE, so a 500 response instead of 400. I feel that is probably OK bc a missing name is more of an internal error rather than a problem with the request.

bryanck:
LGTM, just some nits

bryanck (approved):

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
gabeiglio: Fixed the NPE risk. On the other hand, at that point the alias is already resolved, I am just porting some logic from MetacatController into TableServiceImpl, specifically this code
```java
    Preconditions.checkArgument(table.getName() != null
            && tableName.equalsIgnoreCase(table.getName().getTableName()
        ),
        "Table name does not match the name in the table"
    );
```
Since this code would run after resolving the alias in the controller

metacat-main/src/main/java/com/netflix/metacat/main/services/impl/TableServiceImpl.java
gabeiglio: fixed, isAlias performs a isDisabled check

PR #715, opened 2026-08-11
remove block for secure view load

ldsantos0911:
Currently, there is a block in the PolarisTableConnector get path to prevent plainly loading secure views. I'm going to be adding internal logic to do a partial load and return the Schema in a follow-up [PR](https://github.netflix.net/corp/dseplat-metacat-netflix/pull/2016), so I'm unblocking the load path here.


ursetta-netflix (approved):

yingjianwu98 (approved):

yingjianwu98:
Changes lgtm, and if we cannot get the new service done before the deadline, we can merge this.

PR #721, opened 2026-09-06
Add metadata hash column and store null

bryanck:
This PR adds a new column to store a hash of the Iceberg metadata. This is for compatibility with IRC which will store the hash. Metacat will store `null` to overwrite any previously stored hash from IRC, to prevent the hash from becoming stale. IRC will skip the hash check if set to null. (The end goal is to deprecate Metacat and move all table operations to IRC with the hash check.)


gabeiglio:
This is for securing metadata when performing a stage create right?

gabeiglio (approved):

bryanck:
> This is for securing metadata when performing a stage create right?

This for compatibility with IRC's metadata tamper detection.

yingjianwu98 (approved):

PR #724, opened 2026-09-12
Expose schema to field DTO conversion and add view DTO

jfz:


metacat-iceberg-dto/src/main/java/com/netflix/metacat/iceberg/IcebergTableDtoFactory.java
```diff
@@ -176,6 +223,21 @@ private static Map<String, String> toParams(final String metadataLocation, final
...
+    private static StorageDto toViewSerde(final String viewLocation) {
+        final StorageDto serde = new StorageDto();
+        serde.setUri(viewLocation);
+        return serde;
+    }
+
+    private static Map<String, String> toViewParams(final String metadataLocation,
+                                                    final Map<String, String> properties) {
+        final Map<String, String> params = new HashMap<>(properties);
+        params.remove(PREVIOUS_METADATA_LOCATION_PROP);
```
yingjianwu98: hmm, curious why do we remove
PREVIOUS_METADATA_LOCATION_PROP?

Haven't checked the code but won't it cause issues for commit as we need the previous metadata location?

yingjianwu98:
Putting a note here, can we have our internal logic reuse the same functions? I would prefer not have duplicate code in two places

yingjianwu98 (approved):

